PR #1, opened 2020-04-28
Add hidden option url-template

vandot:
You have to provide custom url-template for github actions and local testing 
If url-template is provided namespace is not needed

### DIFF
--- a/cmd/beekeeper/cmd/check_peercount.go
+++ b/cmd/beekeeper/cmd/check_peercount.go
@@ -43,7 +43,7 @@ Expected peer count equals: node-count + bootnode-count - 1.`,
 	cmd.Flags().IntP(optionNameNodeCount, "c", 1, "node count")
 	cmd.Flags().StringP(optionNameNamespace, "n", "", "Kubernetes namespace")
 	cmd.Flags().StringP(optionNameURLTemplate, "u", "", "URL template")
-	if err := cmd.Flags().MarkHidden("url-template"); err != nil {
+	if err := cmd.Flags().MarkHidden(optionNameURLTemplate); err != nil {
 		panic(err)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/beekeeper/cmd/check_peercount.go
```diff
@@ -25,17 +26,24 @@ Expected peer count equals: node-count + bootnode-count - 1.`,
...
 			return c.config.BindPFlags(cmd.Flags())
 		},
 	}
 
 	cmd.Flags().IntP(optionNameBootNodeCount, "b", 1, "bootnode count")
 	cmd.Flags().IntP(optionNameNodeCount, "c", 1, "node count")
 	cmd.Flags().StringP(optionNameNamespace, "n", "", "Kubernetes namespace")
-	if err := cmd.MarkFlagRequired(optionNameNamespace); err != nil {
+	cmd.Flags().StringP(optionNameURLTemplate, "u", "", "URL template")
+	if err := cmd.Flags().MarkHidden("url-template"); err != nil {
```
janos, line 46: Use variable `optionNameURLTemplate ` instead the explicit value.

pkg/check/peercount.go
```diff
@@ -25,7 +26,13 @@ func PeerCount(opts PeerCountOptions) (err error) {
 	var expectedPeerCount = opts.NodeCount + opts.BootNodeCount - 1
 
 	for i := 0; i < opts.NodeCount; i++ {
-		nodeURL, err := url.Parse(fmt.Sprintf(debugAPITemplateURL, i, opts.Namespace))
+		var nodeURL *url.URL
+		var err error
+		if opts.URLTemplate != "" {
+			nodeURL, err = url.Parse(fmt.Sprintf(opts.URLTemplate, i))
```
janos, line 32: This Sprintf is assuming that the url template will not have a namespace in it. I am suggesting to use template package to format the url instead Sprintf. This can be done in a separate PR as it extends its scope, but this implementation should not stay for long.

PR #20, opened 2020-06-18
pullsync integration tests

acud:


pkg/bee/cluster.go
```diff
@@ -110,6 +110,57 @@ func (c *Cluster) AddressesStream(ctx context.Context) <-chan AddressesStreamMsg
 	return addressStream
 }
 
+// GlobalReplicationFactor returns replication factor for a given chunk
```
janos, line 113: Just noticing that this is not used. It is good to have this method.

PR #22, opened 2020-06-21
Chaos mesh operator

vandot:
Enable chaos test scenarios


cwen0:
Hi, I am the maintainer of Chaos Mesh. It is very nice to see Chaos Mesh being used to test your application. If you have any trouble using Chaos Mesh. please file an [issue](https://github.com/pingcap/chaos-mesh/issues/new/choose) and discuss it with us.  Best wishes!

zhouqiang-cl:
Hi， @vandot I see in this PR，you use `chaos-mesh-controller-manager` directly and not use schedule， may I ask why(maybe we should enhance Chaos Mesh® to meet more scenario)， In recently version(master), Chaos Mesh® also provide API to trigger Chaos, which may be also useful

vandot:
Hi @zhouqiang-cl not sure what do you mean with use directly and not schedule.
We have a need for a easy and dynamic way to trigger new chaos scenarios, thats why CRDs are "templated" and used in that way.
`beekeeper` creates new CRD using k8s api and `chaos-mesh-operator` does the rest.

zhouqiang-cl:
@vandot  Thank you for your explanation, I think "beekeeper creates new CRD using k8s api and chaos-mesh-operator does the rest." already answer "you use chaos-mesh-controller-manager directly" 😃

Can I add swarm or beekeeper as a Chaos Mesh adoptor [here](https://github.com/pingcap/chaos-mesh/blob/master/ADOPTERS.md) ?

zhouqiang-cl:
hi, @vandot sorry to disturb you again. Can I add swarm or beekeeper as a Chaos Mesh adoptor in https://github.com/pingcap/chaos-mesh/blob/master/ADOPTERS.md ?

vandot:
Hi @zhouqiang-cl sorry for the late reply, team doesn't have anything against it, we would be glab to be listed as adopters but only after we start using it.
Sorry, we were busy with alpha release and actually we are still not using chaos-mesh for testing, its only implemented.
If we start using it I will ping you.
Once more sorry for the late reply.

zhouqiang-cl:
@vandot That's ok, thank you very much, If you have any trouble in using Chaos Mesh, please also ping us or file an issue in Chaos Mesh

PR #23, opened 2020-06-21
Helm support

vandot:
Initial helm upgrade support


vandot:
@svetomir lint is passing on my machine, here it says `level=error msg="Timeout exceeded: try increasing it by passing --timeout option"` have you experienced that issue?

PR #25, opened 2020-06-23
Feature/metrics

vandot:


cmd/beekeeper/cmd/check.go
```diff
@@ -47,8 +50,10 @@ func (c *command) initCheckCmd() (err error) {
 	cmd.PersistentFlags().BoolVar(&insecureTLSDebugAPI, optionNameDebugAPIInsecureTLS, false, "skips TLS verification for debug API")
 	cmd.PersistentFlags().BoolVar(&disableNamespace, optionNameDisableNamespace, false, "disable Kubernetes namespace")
 	cmd.PersistentFlags().Bool(optionNameInsecureTLS, false, "skips TLS verification for both API and debug API")
-	cmd.PersistentFlags().StringP(optionNameNamespace, "n", "", "Kubernetes namespace, must be set or disabled")
+	cmd.PersistentFlags().StringP(optionNameNamespace, "n", "bee", "Kubernetes namespace, must be set or disabled")
```
janos, line 53: Is this dangerous, as it is defaulting to the public cluster namespace?

pkg/check/fileretrieval/fileretrieval.go
```diff
@@ -37,34 +50,65 @@ func Check(c bee.Cluster, o Options) (err error) {
...
+			}
 		}
 	}
 
 	return
 }
 
 // CheckFull uploads files on cluster and downloads them from the all nodes in the cluster
-func CheckFull(c bee.Cluster, o Options) (err error) {
+func CheckFull(c bee.Cluster, o Options, pusher *push.Pusher, metrics bool) (err error) {
```
janos, line 96: Is metrics bool redundant? If pusher is not nil, then the metrics are pushed, otherwise not.

cmd/beekeeper/cmd/check.go
```diff
@@ -47,8 +50,10 @@ func (c *command) initCheckCmd() (err error) {
...
 	cmd.PersistentFlags().IntP(optionNameNodeCount, "c", 1, "node count")
+	cmd.PersistentFlags().String(optionNamePushGateway, "http://localhost:9091/", "Prometheus PushGateway")
+	cmd.PersistentFlags().BoolVar(&enableMetrics, optionNameMetrics, false, "send metrics to pushgateway")
```
janos, line 56: Does it make sense to call this `push-metrics` instead `enable-metrics`?

cmd/beekeeper/cmd/check_fileretrieval.go
```diff
@@ -43,13 +44,16 @@ and attempts retrieval of those files from the last node in the cluster.`,
...
 				DebugAPIInsecureTLS:     insecureTLSDebugAPI,
+				DisableNamespace:        disableNamespace,
 				Namespace:               c.config.GetString(optionNameNamespace),
 				Size:                    c.config.GetInt(optionNameNodeCount),
 			})
 			if err != nil {
 				return err
 			}
 
+			pusher := push.New(c.config.GetString(optionNamePushGateway), c.config.GetString(optionNameNamespace))
```
janos, line 55: Pushed does not have to constructed if metrics are not supposed to be pushed.

cmd/beekeeper/cmd/check.go
vandot: @janos have some other suggestion? `default`?

janos: Why not to leave it blank as it is now?

pkg/check/fileretrieval/fileretrieval.go
vandot: `pusher` is used in too many places, it was quicker to set metrics

cmd/beekeeper/cmd/check_fileretrieval.go
vandot: as stated [here](https://github.com/ethersphere/beekeeper/pull/25#discussion_r444263520) `pusher` is used in too many places, after release we will refactor what is needed

janos: ok

PR #30, opened 2020-07-28
change fileretrievaldynamic to NetworkPartition

vandot:


vandot:
@svetomir this PR is not good, discovered bug in chaos-mesh for NetworkChaos, it can be only merged after that bug is fixed upstream

zhouqiang-cl:
> @svetomir this PR is not good, discovered bug in chaos-mesh for NetworkChaos, it can be only merged after that bug is fixed upstream

Can you give a detail bug of chaos-mesh or file an issue in [here](https://github.com/chaos-mesh/chaos-mesh/issues/new/choose)? We will look into it and fix as soon as possible

zhouqiang-cl:
Since Chaos Mesh have already join CNCF sandbox, we need to move it to chaos-mesh org, If update Chaos Mesh, please also update apiVersion to chaos-mesh.org/v1alpha1

zhouqiang-cl:
We have refine the network partition, could you please update chaos mesh and test it again? @vandot 
Thank you very much

vandot:
tested and commented here https://github.com/chaos-mesh/chaos-mesh/issues/763#issuecomment-671404255

PR #32, opened 2020-07-28
finalize pullsync integration tests

acud:


pkg/bee/cluster.go
```diff
@@ -142,7 +142,7 @@ func (c *Cluster) AddressesStream(ctx context.Context) <-chan AddressesStreamMsg
 }
 
 // GlobalReplicationFactor returns replication factor for a given chunk
-func (c *Cluster) GlobalReplicationFactor(ctx context.Context, a swarm.Address) (gcrf float64, err error) {
+func (c *Cluster) GlobalReplicationFactor(ctx context.Context, a swarm.Address) (int, error) {
```
janos, line 145: This function function now returns the total number of nodes that contain a particular chunk. The term factor should be described in the comment with this change in function semantics.

PR #36, opened 2020-08-14
add retry for pullsync check

vandot:


pkg/check/pullsync/pullsync.go
```diff
@@ -132,17 +132,25 @@ func Check(c bee.Cluster, o Options) (err error) {
...
+				t := 1
+				for {
+					time.Sleep(2 * time.Duration(t) * time.Second)
+					synced, err := c.Nodes[ni].HasChunk(ctx, chunk.Address())
+					if err != nil {
+						return fmt.Errorf("node %d: %w", ni, err)
+					}
+					if synced {
+						break
+					} else {
```
janos, line 147: `else` is not needed because of the `break` in the if.

PR #40, opened 2020-08-19
Chunk repair testcase support

jmozah:
This PR will add the support for testing chunk repair alias global pinning in the cluster. It should do the following
- Add chunk remove api support
- create command for chunk reapir
- create the check for chunk repair
- test in the cluster

### DIFF
--- a/cmd/beekeeper/cmd/check_chunkrepair.go
+++ b/cmd/beekeeper/cmd/check_chunkrepair.go
@@ -55,7 +55,7 @@ and attempts repairing of those chunks for the other nodes in the cluster.`,
 			}
 
 			return chunkrepair.Check(cluster, chunkrepair.Options{
-				NumberOfChunksToRepair: c.config.GetInt(optionNameChunkRepairNodeCount),
+				NumberOfChunksToRepair: c.config.GetInt(optionNumberOfChunks),
 				Seed:                   seed,
 			}, pusher, c.config.GetBool(optionNamePushMetrics))
 
--- a/cmd/beekeeper/cmd/check_chunkrepair.go
+++ b/cmd/beekeeper/cmd/check_chunkrepair.go
@@ -63,7 +63,7 @@ and attempts repairing of those chunks for the other nodes in the cluster.`,
 		PreRunE: c.checkPreRunE,
 	}
 
-	cmd.Flags().IntP(optionNameChunkRepairNodeCount, "rc", 3, "number of nodes required to run the chunk repair test")
+	cmd.Flags().IntP(optionNameChunkRepairNodeCount, "r", 3, "number of nodes required to run the chunk repair test")
 	cmd.Flags().Int64P(optionNameSeed, "s", 0, "seed for generating chunks; if not set, will be random")
 	cmd.Flags().Float64(optionNumberOfChunks, 1, "no of chunks to repair")
 	return cmd
--- a/cmd/beekeeper/cmd/check_chunkrepair.go
+++ b/cmd/beekeeper/cmd/check_chunkrepair.go
@@ -1,0 +1,70 @@
+package cmd
+
+import (
+	"errors"
+
+	"github.com/ethersphere/beekeeper/pkg/bee"
+	"github.com/ethersphere/beekeeper/pkg/check/chunkrepair"
+	"github.com/ethersphere/beekeeper/pkg/random"
+	"github.com/prometheus/client_golang/prometheus/push"
+	"github.com/spf13/cobra"
+)
+
+func (c *command) initCheckChunkRepair() *cobra.Command {
+	const (
+		optionNameChunkRepairNodeCount = "chunk-repair-node-count"
+		optionNameSeed                 = "seed"
+		optionNumberOfChunks           = "number-of-chunks"
+	)
+
+	cmd := &cobra.Command{
+		Use:   "chunkrepair",
+		Short: "Checks chunk repair ability of the cluster",
+		Long: `Checks chunk repair ability of the cluster.
+It uploads given number of chunks to given number of nodes, 
+and attempts repairing of those chunks for the other nodes in the cluster.`,
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if c.config.GetInt(optionNameChunkRepairNodeCount) > c.config.GetInt(optionNameNodeCount) {
+				return errors.New("bad parameters: chunk-repair-node-count must be less or equal to node-count")
+			}
+
+			cluster, err := bee.NewCluster(bee.ClusterOptions{
+				APIScheme:               c.config.GetString(optionNameAPIScheme),
+				APIHostnamePattern:      c.config.GetString(optionNameAPIHostnamePattern),
+				APIDomain:               c.config.GetString(optionNameAPIDomain),
+				APIInsecureTLS:          insecureTLSAPI,
+				DebugAPIScheme:          c.config.GetString(optionNameDebugAPIScheme),
+				DebugAPIHostnamePattern: c.config.GetString(optionNameDebugAPIHostnamePattern),
+				DebugAPIDomain:          c.config.GetString(optionNameDebugAPIDomain),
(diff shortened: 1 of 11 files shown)


pkg/bee/node.go
```diff
@@ -316,6 +316,15 @@ func (n *Node) UploadChunk(ctx context.Context, c *Chunk) (err error) {
 	return
 }
 
+// RemoveChunk removes the given chunk from the node's local store
+func (n *Node) RemoveChunk(ctx context.Context, c *Chunk) (err error) {
+	err = n.debug.Chunks.Remove(ctx, c.Address())
+	if err != nil {
+		return fmt.Errorf("remove chunk: %w", err)
+	}
+	return
```
janos, line 325:
```suggestion
	return n.debug.Chunks.Remove(ctx, c.Address())
```

pkg/beeclient/api/api.go
```diff
@@ -185,6 +185,9 @@ func drain(r io.ReadCloser) {
 // responseErrorHandler returns an error based on the HTTP status code or nil if
 // the status code is from 200 to 299.
 func responseErrorHandler(r *http.Response) (err error) {
+	if r.StatusCode == http.StatusAccepted {
+		return ErrRecoveryInitiated
+	}
```
janos, line 190: Isn't this too specific to a global pinning feature in order to be in used in every request? Status accepted is not a universal equivalence of recovery in the api. I think that this status should be handled specifically only in cases where chunk repair feature exists.

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+			return err
+		}
+
+		for {
+			// check if the node is there in the local store of node B
+			// this does a get chunk instead of Has chunk, so the following
+			// call just checks if the chunk is accessible from nodeB
+			present, err := nodeB.HasChunk(ctx, chunk.Address())
+			if err != nil {
+				fmt.Println("sleeping...")
```
janos, line 60: This could print a lot of `sleeping...` in terminal, very fast. Could this retry be improved and limited?

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// findFarthestNodes finds two farthest nodes in the cluster
+func findFarthestNodes(overlays []swarm.Address) (swarm.Address, swarm.Address, error) {
+	var overlayA swarm.Address
+	var overlayC swarm.Address
+	dist := big.NewInt(0)
+	for _, a := range overlays {
+		for _, c := range overlays {
+			if bytes.Equal(a.Bytes(), c.Bytes()) {
```
janos, line 241: Use swarm.Address.Equal.

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+	err := node.UploadChunk(ctx, chunk)
+	if err != nil {
+		return err
+	}
+	pinned, err := node.PinChunk(ctx, chunk.Address())
+	if err != nil {
+		return err
+	}
+	if !pinned {
+		return errors.New("could not pin chunk in nodeA")
```
janos, line 200: `nodeA` is too specific for this function, it can be removed from the error message.

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+		// it is hard to simulate the chunk failure in small clusters. We would need a
+		// fairly large cluster then.
+		err = deleteChunkFromAllNodes(ctx, c, chunk)
+		if err != nil {
+			return err
+		}
+
+		// trigger downloading of the chunk from nodeC again (this time it should trigger chunk repair)
+		_, err = nodeC.DownloadChunk(ctx, chunk.Address(), addressA.String()[0:2])
+		if err == nil { // for status Accepted (209), a error is returned
```
janos, line 91: An error condition is when the error value is nil, but in that case, the nil error is returned.

Check for the expected error and return an appropriate error stating what is the problem, instead just nil.

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+			return err
+		}
+
+		// by the time the NodeC creates a trojan chunk and asks NodeA to repair, upload the
+		// original chunk in nodeA and pin it
+		err = uploadAndPinChunkToNode(ctx, &nodeA, chunk)
+		if err != nil {
+			return err
+		}
+		time.Sleep(5 * time.Second) // give sometime so that the repair happens
```
janos, line 101: This is an arbitrary wait. Would it be better to have a retry here?

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil { // this time it should succeed
+			return err
+		}
+		if !bytes.Equal(data3, chunk.Data()) {
+			return errors.New("chunk downloaded in NodeC does not have proper data")
+		}
+		fmt.Println("repaired chunk ", chunk.Address().String())
+
+		if pushMetrics {
+			if err := pusher.Push(); err != nil {
```
janos, line 114: None of the metrics are changed during this test. Basically they are not used. This push call will always push the same initial values of all metrics.

cmd/beekeeper/cmd/check_chunkrepair.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/ethersphere/beekeeper/pkg/bee"
+	"github.com/ethersphere/beekeeper/pkg/check/chunkrepair"
+	"github.com/ethersphere/beekeeper/pkg/random"
+	"github.com/prometheus/client_golang/prometheus/push"
+	"github.com/spf13/cobra"
+)
+
+func (c *command) initCheckChunkRepair() *cobra.Command {
+	const (
+		optionNameChunkRepairNodeCount = "chunk-repair-node-count"
```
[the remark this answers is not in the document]
jmozah: it is used only in the checking

pkg/beeclient/api/api.go
jmozah: you are right. But i could not find any other way to check if global pinning is triggered. I thought you will not get any data when you send 202.. but i get some data... not sure from where... Let me know how this can be handled.

pkg/check/chunkrepair/chunkrepair.go
jmozah: not much for a cluster of our size... but i think this print is superfluous and can be removed.

cmd/beekeeper/cmd/check_chunkrepair.go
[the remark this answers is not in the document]
janos: But why? The check itself does not use this information at all.

pkg/check/chunkrepair/chunkrepair.go
janos: Well, this can be an endless loop regardless of the print.

cmd/beekeeper/cmd/check_chunkrepair.go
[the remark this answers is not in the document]
jmozah: It is not checked in "check" as it is already checked inside "cmd/check_chunkrepair.go".
if you want i can add a redundant check inside pkg/check/chunkrepair/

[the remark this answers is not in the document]
janos: I do not see the need for cli flag just to check if another cli flag is set. It is possible to provide `--chunk-repair-node-count 1` and have not functional test. Making sure that at least 3 nodes are started does not require a cli flag.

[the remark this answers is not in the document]
jmozah: removed this flag

janos (approved):

svetomir (approved):

PR #46, opened 2020-09-11
instant chunk repair

jmozah:
Pick 3 nodes `NodeA`, `NodeB`, `NodeC` and a chunk `ch`, such that

`NodeA` is farthest from `NodeC` in the cluster
chunk address of `ch` is the closest to `NodeB`

- Upload and pin the chunk in `NodeA`
- Check if it reached `NodeB`
- Download the chunk from NodeC
- Delete chunk from all nodes except `NodeA`
- Now try downloading with `NodeA` as target (this time we should get the repaired chunk)
- Download the chunk again without target (it should succeed now as chunk repairing would have placed the chunk in proper node in the network)

### DIFF
--- a/pkg/check/chunkrepair/chunkrepair.go
+++ b/pkg/check/chunkrepair/chunkrepair.go
@@ -51,12 +51,12 @@ func Check(c bee.Cluster, o Options, pusher *push.Pusher, pushMetrics bool) (err
 	}
 
 	// for every chunk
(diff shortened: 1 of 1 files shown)


svetomir (approved):

acud:
closing this as it is related to the instant chunk repair that was put on hold for now.

related to https://github.com/ethersphere/bee/pull/676

PR #51, opened 2020-09-18
Settlements

svetomir:


pkg/check/balances/balances.go
```diff
@@ -40,6 +40,7 @@ func Check(c bee.Cluster, o Options, pusher *push.Pusher, pushMetrics bool) (err
 		if err := validateBalances(overlays, b); err != nil {
 			return fmt.Errorf("invalid balances before uploading a file")
 		}
+		fmt.Printf("Valid balances\n")
```
janos, line 43: fmt.Println? in three places below, also, not so important.

janos (approved):
Cool, thanks @svetomir.

PR #55, opened 2020-09-23
initial suite.sh script

vandot:
Initial version of suite.sh script
Note that it has relative path to beekeeper binary


janos (approved):
Very cool! Thanks @vandot.

svetomir (approved):

PR #57, opened 2020-10-14
clean up existing tests, rewrite local pinning tests

acud:
This PR involved originally cleaning up the local pinning tests, but gradually grew to be a bit of a larger refactor.
Changes are namely:
- remove `bee.Chunk` in favor of `swarm.Chunk`
- Create a chunk not found test vector to see that GC is working correctly - upload one chunk (no pinning), then fill the store and make sure it is collected. This check is accessible under `beekeeper check gc`
- Rewrite the local pinning tests to be conclusive - added a check for one chunk pinned manually, and also a check that an arbitrary byte size is also pinned (all chunks are asserted to exist on the pinning node)
- Add `pipeline` dependency from `bee` to determine chunks in an arbitrary size data before sending to the node
- Fixed a lot of godocs
- Added bmtpool from `bee` to get rid of manual bmt hashing across the codebase
- Cleaned up the chunk helpers to reduce coupling
- Resolved some cases where value semantics and pointer semantics usage was confused and not clear
- Removed metrics from local pinning tests. Not sure if we need them at all, and I am not really sure which sort of metrics we need there. With that in mind, it seems we are copy pasting a lot of metrics code in the repo and kind of blindly just copy-pasting stuff without really using it in grafana, which I am not so big fan of. So for now I opted out of metrics in local pinning and gc tests. If this is necessary we can add them.


pkg/bee/chunk.go
```diff
@@ -3,90 +3,41 @@ package bee
...
 )
 
-const (
-	// MaxChunkSize represents max chunk size in bytes
-	MaxChunkSize = 4096
-	spanInfoSize = 8
-)
-
-// Chunk represents Bee chunk
-type Chunk struct {
```
svetomir, line 21: This type represents this pkg's type which has slightly different purpose and usage than the swarm.Chunk type.
In the future it can contain some additional fields needed for beekeeper.
It can be used as a wrapper around swarm.Chunk type, i.e. swarm.Chunk can be one filed in this struct.
Logic of the whole pkg is designed like that - this way underlaying technologies can be easily managed, new can be added or old ones can be even replaced.

What was the reason for replacing it with the other pkg's type?

svetomir (commented):
@acud @janos I reviewed it to the level of beekeeper logic and implementation.
@janos Can you please review testing logic from bee point of view?

pkg/bee/chunk.go
acud: the main problem is that it is not a 1:1 implementation. in some details it is incorrectly enforcing some constraints such as size (maximum data part of the chunk is always 4096, span is appended to that, but in the implementation here the data of the chunk is 4096 minus span, which is incorrect), and in other cases it separates span and size, which is incorrect since they are in normal swarm chunks not separate from the data. i found it a bit confusing to maintain a parallel implementation that is no 1:1, it also prevents interoperability with helper functions and other component in `bee` (which is a dependency).

janos: I think that the original idea was that Chunk in bee and chunk in beekeeper have different purposes and that beekeeper requirements do not have to be implemented in bee. The correction in data size should be trivial without so much changes.

I will give my review, but this PR has both architecture refactoring and new and changed checks with a bit larger number of lines changed. It would be better to have PRs that are focused on one purpose, if possible.

svetomir: Would this be helpful?

`type Chunk struct {
	Swarm swarm.Chunk
}`

pkg/beeclient/api/bytes.go
```diff
@@ -22,7 +22,11 @@ type BytesUploadResponse struct {
 }
 
 // Upload uploads bytes to the node
-func (b *BytesService) Upload(ctx context.Context, data io.Reader) (resp BytesUploadResponse, err error) {
-	err = b.client.request(ctx, http.MethodPost, "/"+apiVersion+"/bytes", data, &resp)
+func (b *BytesService) Upload(ctx context.Context, data io.Reader, pin bool) (resp BytesUploadResponse, err error) {
```
janos, line 25: Pin argument is optional and it would be better to have it as the part of the options struct passed to this method as the last argument. This allows other options (like tag and encrypt) to be easily added instead to change the function signature.

pkg/check/gc/chunk_notfound.go
```diff
@@ -0,0 +1,67 @@
+package gc
```
janos, line 1: Add copyright header.

pkg/check/localpinning/bytes_found.go
```diff
@@ -0,0 +1,82 @@
+package localpinning
```
janos, line 1: Add copyright header.

pkg/check/localpinning/chunk_found.go
```diff
@@ -0,0 +1,68 @@
+package localpinning
```
janos, line 1: Add copyright header.

pkg/check/localpinning/metrics.go
```diff
@@ -1,114 +0,0 @@
-package localpinning
```
janos, line 1: Should there be some other metrics added instead these ones.

janos (commented):
New checks have more correct logic and they are better. But this PR mixes both refactoring and check development and also some improvements in comments that would have been better to have in different PRs.

I am a bit concerned about tiding bee and beekeeper too much directly.

cmd/beekeeper/cmd/check_localpinning.go
```diff
@@ -1,42 +1,69 @@
...
 	"github.com/ethersphere/beekeeper/pkg/bee"
 	"github.com/ethersphere/beekeeper/pkg/check/localpinning"
 	"github.com/ethersphere/beekeeper/pkg/random"
-	"github.com/prometheus/client_golang/prometheus/push"
 
 	"github.com/spf13/cobra"
 )
 
-func (c *command) initCheckLocalPinning() *cobra.Command {
+func (c *command) initCheckLocalPinning() []*cobra.Command {
```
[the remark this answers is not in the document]
acud: there are multiple checks for local pinning and the flags are shared, why to overload them into different files?

pkg/bee/node.go
```diff
@@ -139,12 +140,33 @@ func (n *Node) DownloadFile(ctx context.Context, a swarm.Address) (size int64, h
...
-// Overlay returns node's overlay address
+// HasChunkRetry checks if a node has a chunk and retries if not.
+func (n *Node) HasChunkRetry(ctx context.Context, a swarm.Address, retries int) (bool, error) {
+	for i := 0; i <= retries; i++ {
+		has, err := n.HasChunk(ctx, a)
+		if err != nil {
+			if i == retries {
+				return false, err
+			}
+			if errors.Is(err, api.ErrServiceUnavailable) {
```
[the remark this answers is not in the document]
acud: the github tests were failing because of this. this is not new in the codebase, just swapped locations

pkg/check/gc/chunk_notfound.go
acud: there are no copyright headers across the entire repo (at all). should i really do this as part of this pr?

pkg/check/localpinning/metrics.go
acud: AFAIK we dont use these metrics. i also wrote this in the PR description. @svetomir can you confirm if have grafana dashboards for localpinning test? if we don't I rather not spend time on this

pkg/bee/chunk.go
acud: > I think that the original idea was that Chunk in bee and chunk in beekeeper have different purposes and that beekeeper requirements do not have to be implemented in bee. The correction in data size should be trivial without so much changes.

I thought that you gave your consent to do this change, we discussed this no? Maybe I got you wrong.
OK.
You want the old chunk type back?

cmd/beekeeper/cmd/check_localpinning.go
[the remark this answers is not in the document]
janos: For consistency and clarity.

pkg/bee/node.go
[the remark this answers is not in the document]
janos: It deserves at least a comment.

pkg/check/gc/chunk_notfound.go
janos: Ah, sorry, I was reviewing with bee in mind.

janos:
Why is this pr closed?

acud:
I did not realise that I am trying to merge so much contentions content in this PR. I will split it into a PR only for local pinning tests. Then if you want to keep the old chunk type or new one we can discuss but this seems to be not in line with what you guys want so I will revert it

janos:
Thanks for the explanation. I think this PR can be adjusted according to the discussion that is here. I did not require the split. It is just a comment for the future, not to prolong merging of changes in this one.

acud:
it was easier to just start from a new branch.

PR #62, opened 2020-11-03
Add pushsync check for files using tags

zbiljic:
Introduces additional check for 'pushsync' that uploads files to the node (with tag configured), and then checks if counters from tag are expected.
This change also adds `TagsService`, and accompanying functions, to allow using that API.

From the tag counters, `split` value is used as expected number of chunks that need to be pushed, `sent` value is used to see how many chunks have been pushed so far, and `seen` value as number of chunks that are left on the uploading node, and not synced (due to the distribution of chunks through the cluster). For this `seen` value, I was not aware, but have observed on last testing session (generally was always `0`).

In the 'normal' operation, these numbers should usually match, otherwise it might indicate there is some problem.

These changes have been tested on some previous Bee version (before one revert regarding this), and has shown 10-15% more synced chunks.

relates to: [ethersphere/bee#885](https://github.com/ethersphere/bee/issues/885)

### DIFF
--- a/cmd/beekeeper/cmd/check_pushsync.go
+++ b/cmd/beekeeper/cmd/check_pushsync.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"errors"
+	"time"
 
 	"github.com/ethersphere/beekeeper/pkg/bee"
 	"github.com/ethersphere/beekeeper/pkg/check/pushsync"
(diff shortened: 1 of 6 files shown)


janos (approved):
👍

svetomir (approved):

PR #67, opened 2020-11-13
Fix flaky tests

acud:
Identified a bug where we were uploading a chunk to the bytes api. Because when generating the chunk, the span is appended to the random generated bytes, the hash which was returned from the bytes API was different than the hash generated in the beekeeper tests, since bmt hash is calculated without the span in the actual data being hashed, and the bytes api treated the span data that was appended to the data as actual data, not span data, resulting in a different hash. So actually the data chunk that was uploaded had a different hash, but the checks inside the actual tests use `chunk.ClosestNode`, which correlates to the correct hash (which wasn't in fact uploaded correctly), resulting in incorrect assumptions, and chunks which are never found where they are supposed to be.

### DIFF
--- a/pkg/bee/chunk.go
+++ b/pkg/bee/chunk.go
@@ -33,9 +33,18 @@ func NewChunk(data []byte) (Chunk, error) {
 	return Chunk{data: data}, nil
 }
 
+// NewChunk returns new chunk
+func NewChunkA(data []byte, a swarm.Address) (Chunk, error) {
+	if len(data) > MaxChunkSize {
+		return Chunk{}, fmt.Errorf("create chunk: requested size too big (max %d bytes)", MaxChunkSize)
+	}
+
+	return Chunk{data: data, address: a}, nil
+}
+
 // NewRandomChunk returns new pseudorandom chunk
-func NewRandomChunk(r *rand.Rand) (c Chunk, err error) {
-	data := make([]byte, r.Intn(MaxChunkSize-spanInfoSize))
+func NewRandomChunk(r *rand.Rand) (Chunk, error) {
+	data := make([]byte, r.Intn(MaxChunkSize))
 	if _, err := r.Read(data); err != nil {
 		return Chunk{}, fmt.Errorf("create random chunk: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


janos (commented):
This PR makes tests more semantically correct by using UploadChunk instead UploadBytes. But UploadBytes is just uploading an arbitrary data, that is passed from chunk.Data() and the address that is returned is correct as it is the same address as returned by bee.

I do not see that this PR fixed the problem about addresses that is described, as in this PR all changes are related to the address provided by the bee bytes api, non of the addresses are from beekeeper's Chunk.Address().

Another approach to this PR would be just to generate some random data (<=4096 bytes) and pass to UploadBytes, instead to use chunk.Data() to get it, which would be also more semantically correct in regards to UploadBytes.

It can be assumed that with such bug that is described, the flakiness would be much higher even checks not functional.

acud:
> I do not see that this PR fixed the problem about addresses that is described

I would kindly ask you to read the tests again.
The address coming from the chunk is used, but it is not correct, since the data that was actually uploaded results in a different content address hash.

Example:
https://github.com/ethersphere/beekeeper/blob/master/pkg/check/pullsync/pullsync.go#L94
https://github.com/ethersphere/beekeeper/blob/master/pkg/check/pushsync/pushsync.go#L70
in both cases the actual chunk that was uploaded is the address which is returned from `UploadBytes`, but the `ClosestNode` compares the original chunk address (which is correct, but it is not the real content addressed chunk that was uploaded to the cluster)

janos:
> > I do not see that this PR fixed the problem about addresses that is described
> 
> I would kindly ask you to read the tests again.
> The address coming from the chunk is used, but it is not correct, since the data that was actually uploaded results in a different content address hash.
> 
> Example:
> https://github.com/ethersphere/beekeeper/blob/master/pkg/check/pullsync/pullsync.go#L94
> https://github.com/ethersphere/beekeeper/blob/master/pkg/check/pushsync/pushsync.go#L70
> in both cases the actual chunk that was uploaded is the address which is returned from `UploadBytes`, but the `ClosestNode` compares the original chunk address (which is correct, but it is not the real content addressed chunk that was uploaded to the cluster)

Yes, you are right.

svetomir (approved):

janos (approved):

PR #74, opened 2020-12-11
Add test for pinning that fetches remote chunks

zbiljic:
relates to: [ethersphere/bee#571](https://github.com/ethersphere/bee/issues/571)

Adds `pin-remote` check that performs pinning of bytes upload and then checks all those chunks are present locally using debug API.

One partly unfortunate thing is that because of how the Bee operates, if there is too few nodes in the topology, it is possible that all of them will replicate all chunks between themselves. As an attempt to "combat" this, few changes were made:
1. number of chunks that are being generated for upload was increased to `8000`
2. timeout to wait for chunks replication is set to `15` seconds
3. there is "verification" step that queries each node for expected chunks, and only on the nodes that do not have all the chunks does it perform pinning
    - initial idea was to return an error if every node has all the chunks, however since that may make this test "flaky" single log line is written to note the fact

### DIFF
--- a/pkg/beeclient/api/pinning.go
+++ b/pkg/beeclient/api/pinning.go
@@ -66,3 +66,105 @@ func (p *PinningService) UnpinChunk(ctx context.Context, a swarm.Address) (bool,
 
 	return true, nil
 }
+
+// PinBytes pins chunks for bytes upload
+func (p *PinningService) PinBytes(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodPost, "/pin/bytes/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// UnpinBytes unpins chunks for bytes upload
+func (p *PinningService) UnpinBytes(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodDelete, "/pin/bytes/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// PinFiles pins chunks for files upload
+func (p *PinningService) PinFiles(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodPost, "/pin/files/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// UnpinFiles unpins chunks for files upload
+func (p *PinningService) UnpinFiles(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodDelete, "/pin/files/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// PinBzz pins chunks for bzz upload
+func (p *PinningService) PinBzz(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodPost, "/pin/bzz/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// UnpinBzz unpins chunks for bzz upload
+func (p *PinningService) UnpinBzz(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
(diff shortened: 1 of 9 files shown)


svetomir (approved):

acud (changes requested):
@zbiljic looks good. I have a few recommendations for simplification:
- returning a bool and an error in the pinning service function signature seems unnecessary to me. i would like to actually avoid using such patterns, since it swallows some errors and translates them into other return values. i would try to phase out of this pattern in the beekeeper codebase slowly. if an error occurs, it is an error and the calling context should decide whether it is an infrastructure error (service unavailable), or a node error (chunk not found), etc. having a bool here does not improve the situation (again, this is only my opinion, but i find this confusing to read and interpret, as some errors are errors, but some aren't)
- about the test itself, the fact that we are relying on nodes that might have or not have certain chunks due to syncing means that we could potentially simplify the test scenario. in general, i think that beekeeper scenarios should be as much as possible under our control from beekeeper, thus, i would consider the following:
- - generate random data
- - get all data chunk references locally before uploading them (reuse `pkg/check/localpinning/localpinning.go#addresses()`)
- - upload the data to one node
- - wait for syncing
- - delete all known data chunks on a random node (debugapi `DELETE /chunks/{addr}`)
- - pin with the traversal api on the node from previous step
- - check same node has all deleted chunks

WDYT?

pkg/beeclient/api/pinning.go
```diff
@@ -57,7 +57,109 @@ func (p *PinningService) UnpinChunk(ctx context.Context, a swarm.Address) (bool,
...
+
+// PinBytes pins chunks for bytes upload
+func (p *PinningService) PinBytes(ctx context.Context, a swarm.Address) (bool, error) {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := p.client.requestJSON(ctx, http.MethodPost, "/pin/bytes/"+a.String(), nil, &resp)
+	if err == ErrNotFound {
```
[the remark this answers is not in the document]
zbiljic: I do not disagree with this comment, these new functions are basically copy of existing one - `PinChunk`, which had it defined this way.

As this projects main thing is testing other project, we cut a lot of corners in terms of "style" for ease of use, which is fine by me, but I can change this.

pkg/beeclient/api/pinning.go
```diff
@@ -57,7 +57,109 @@ func (p *PinningService) UnpinChunk(ctx context.Context, a swarm.Address) (bool,
...
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// UnpinBytes unpins chunks for bytes upload
+func (p *PinningService) UnpinBytes(ctx context.Context, a swarm.Address) (bool, error) {
```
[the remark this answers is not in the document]
zbiljic: Yeah, there is `8` total functions with this response signature.

zbiljic:
In relation to suggested change for the test:

> * about the test itself, the fact that we are relying on nodes that might have or not have certain chunks due to syncing means that we could potentially simplify the test scenario. in general, i think that beekeeper scenarios should be as much as possible under our control from beekeeper, thus, i would consider the following:
> * * generate random data
> * * get all data chunk references locally before uploading them (reuse `pkg/check/localpinning/localpinning.go#addresses()`)
> * * upload the data to one node
> * * wait for syncing
> * * delete all known data chunks on a random node (debugapi `DELETE /chunks/{addr}`)
> * * pin with the traversal api on the node from previous step
> * * check same node has all deleted chunks

the only thing that is currently not being done is deleting chunks from the node, all the other steps are the same.

I have already explained my reasoning for not doing delete through debug API, but it probably should be fine, and I will update test to this.

pkg/beeclient/debugapi/node.go
```diff
@@ -65,6 +65,23 @@ func (n *NodeService) HasChunk(ctx context.Context, a swarm.Address) (bool, erro
...
 }
 
+// RemoveChunk removes chunk from the node
+func (n *NodeService) RemoveChunk(ctx context.Context, a swarm.Address) error {
+	resp := struct {
+		Message string `json:"message,omitempty"`
+		Code    int    `json:"code,omitempty"`
+	}{}
+
+	err := n.client.requestJSON(ctx, http.MethodDelete, "/chunks/"+a.String(), nil, &resp)
```
acud, line 75: `return n.client.request...`

pkg/check/chunkrepair/chunkrepair.go
```diff
@@ -229,13 +229,16 @@ func uploadAndPinChunkToNode(ctx context.Context, node *bee.Client, chunk *bee.C
 	if err != nil {
 		return err
 	}
-	pinned, err := node.PinChunk(ctx, chunk.Address())
+
+	err = node.PinChunk(ctx, chunk.Address())
```
acud, line 233: `return node.PinChunk`

pkg/check/localpinning/remote_chunks_found.go
```diff
@@ -0,0 +1,140 @@
...
+			fmt.Printf("Node %s: pivot node, skipping it\n", name)
+			continue
+		}
+
+		nodeClient := ng.Node(name)
+
+		fmt.Printf("Node %s: removing expected chunks\n", name)
+
+		for _, a := range addrs {
+			err := nodeClient.RemoveChunkWithAddress(ctx, a)
```
acud, line 64: a point to consider, that we might be deleting a chunk which falls under the node's area of responsibility. with a large enough network size this could potentially fail. with a small network it should be OK since the node can still reach the uploader node. just a sidenote, that if we see flakes, this could be the possible reason.

acud (approved):
LGTM, thanks

pkg/check/chunkrepair/chunkrepair.go
zbiljic: I have left this as it was, because I was not sure if it would break existing code, but we will see it if it does I guess (probably should have had separate PR for this change).

pkg/check/localpinning/remote_chunks_found.go
zbiljic: The `DELETE` operation should never fail under normal conditions, even when the node does not have the chunk (why I left code that was commented above). The `RemoveChunkWithAddress` only relates to chunks that are locally present on the node.

At worst case scenario (under normal operations) we should get `InternalServerError` that could happen when the first chunk is checked if it is present locally, and then by the time chunk remove operation starts (i.e. `storage.ModeSetRemove`) the chunk was already removed by some other process. We should probably add check in Bee there in the case when the error was `ErrNotFound` to again just return `OK` response.

acud: I meant that the flake could come up from a certain edge case where we delete all chunks from a node, but some chunks that should be replicated to other nodes did not (due to asymmetric neighborhoods), hence resulting in some chunks being lost forever, then, once you initiate the pinning process, the node might not be able to reach, download and pin some chunks, resulting in a possible flake.

PR #78, opened 2021-01-14
Fix typo in fileHasher function

r0qs:


acud (approved):
Thanks for your contribution @r0qs!

PR #80, opened 2021-01-21
adjust dynamic kademlia check

acud:
@janos @svetomir I think this is something we can start with and iterate on. Would be happy for you feedback.

ref https://github.com/ethersphere/bee/issues/1121

### DIFF
--- a/pkg/check/kademlia/dynamic.go
+++ b/pkg/check/kademlia/dynamic.go
@@ -2,13 +2,14 @@ package kademlia
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/beekeeper/pkg/bee"
-	"github.com/ethersphere/beekeeper/pkg/check/fullconnectivity"
 	"github.com/ethersphere/beekeeper/pkg/random"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/check/kademlia/dynamic.go
```diff
@@ -102,53 +95,105 @@ func CheckDynamic(ctx context.Context, cluster *bee.Cluster, o Options) (err err
...
-
 		topologies, err := cluster.Topologies(ctx)
 		if err != nil {
 			return err
 		}
 
 		fmt.Println("Checking Kademlia")
 		if err := checkKademliaD(topologies); err != nil {
-			fmt.Printf("Kademlia check failed: %v\n", err)
+			fmt.Printf("Dynamic kademlia check failed: %v\n", err)
```
janos, line 105: Is this print needed as the error is retuned?

janos (approved):
LGTM

pkg/check/kademlia/dynamic.go
acud: Will it be printed by cobra? I guess yes right?

svetomir (approved):

svetomir:
@acud Looks good.

And, just a general question - should we remove fullConnectivity from static check also?
If answer is yes, maybe we could do it through this PR.

pkg/check/kademlia/dynamic.go
janos: Yes, looks like that is the case, but actually, the exact caller is not relevant, it can be anything anytime.

acud: :+1:

janos (approved):

PR #83, opened 2021-01-26
bee, check: accommodate for chunk api changes

acud:
needed for https://github.com/ethersphere/bee/pull/1143

### DIFF
--- a/pkg/bee/chunk.go
+++ b/pkg/bee/chunk.go
@@ -25,12 +25,12 @@ type Chunk struct {
 }
 
 // NewChunk returns new chunk
-func NewChunk(data []byte) (Chunk, error) {
-	if len(data) > MaxChunkSize {
-		return Chunk{}, fmt.Errorf("create chunk: requested size too big (max %d bytes)", MaxChunkSize)
+func NewChunk(address swarm.Address, data []byte) (*Chunk, error) {
+	if len(data) > MaxChunkSize+spanInfoSize {
+		return nil, fmt.Errorf("create chunk: requested size too big (max %d bytes)", MaxChunkSize+spanInfoSize)
 	}
 
-	return Chunk{data: data}, nil
+	return &Chunk{address: address, data: data}, nil
 }
 
 // NewRandomChunk returns new pseudorandom chunk
--- a/pkg/bee/chunk.go
+++ b/pkg/bee/chunk.go
@@ -24,15 +24,6 @@ type Chunk struct {
 	span    int
 }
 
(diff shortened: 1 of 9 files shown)


svetomir (approved):

pkg/beeclient/api/chunks.go
```diff
@@ -33,6 +32,6 @@ func (c *ChunksService) Upload(ctx context.Context, a swarm.Address, data io.Rea
 	if o.Pin {
 		h.Add("Swarm-Pin", "true")
 	}
-	err := c.client.requestWithHeader(ctx, http.MethodPost, "/"+apiVersion+"/chunks/"+a.String(), h, data, &resp)
+	err := c.client.requestWithHeader(ctx, http.MethodPost, "/"+apiVersion+"/chunks", h, data, &resp)
```
janos, line 35: It is no longer needed to pass the address to ChunksService.Upload (this) method. I would even say that data as io.Readed could be changed to []byte type as the chunk size is very small, but this is just a minor detail, if you want to consider.

pkg/bee/client.go
```diff
@@ -433,25 +432,13 @@ func (c *Client) UploadBytes(ctx context.Context, b []byte, o api.UploadOptions)
 }
 
 // UploadChunk uploads chunk to the node
-func (c *Client) UploadChunk(ctx context.Context, chunk *Chunk, o api.UploadOptions) (err error) {
-	p := bmtlegacy.NewTreePool(chunkHahser, swarm.Branches, bmtlegacy.PoolSize)
-	hasher := bmtlegacy.New(p)
-	err = hasher.SetSpan(int64(chunk.Span()))
+func (c *Client) UploadChunk(ctx context.Context, chunk *Chunk, o api.UploadOptions) (swarm.Address, error) {
```
janos, line 435: As the chunk address is not needed to Chunks.Upload, this method UploadChunk can just receive chunk data as []byte instead the *Chunk.

pkg/bee/chunk.go
```diff
@@ -25,12 +25,12 @@ type Chunk struct {
 }
 
 // NewChunk returns new chunk
-func NewChunk(data []byte) (Chunk, error) {
-	if len(data) > MaxChunkSize {
-		return Chunk{}, fmt.Errorf("create chunk: requested size too big (max %d bytes)", MaxChunkSize)
+func NewChunk(address swarm.Address, data []byte) (*Chunk, error) {
```
janos, line 28: This constructor does not seem to be used anywhere, consider removing it.

PR #93, opened 2021-02-03
Implement Bee k8s client with interfaces

svetomir:
Hello, this PR implements Bee kubernetes client with interfaces. Goal is to allow better management of pkgs and internal checks. It should be foundation to reorganise pkgs with interfaces. Further improvements will come with additional PRs.


janos (approved):
👍

acud (approved):

PR #102, opened 2021-02-10
Fix dynamic deletion

svetomir:


janos (approved):
One small naming suggestion to rename StartedNodes to RunningNodes. Just for consideration for some other PR.

PR #115, opened 2021-03-22
Stress

svetomir:
Based on https://github.com/ethersphere/bee/issues/1398


pkg/check/uploadstress/uploadstress.go
```diff
@@ -0,0 +1,108 @@
...
+type UploadStress struct{}
+
+// NewUploadStress returns new ping check
+func NewUploadStress() *UploadStress {
+	return &UploadStress{}
+}
+
+// Run executes upload stress check
+func (u *UploadStress) Run(ctx context.Context, cluster *bee.Cluster, o check.Options) (err error) {
+	buffer := 10
```
janos, line 30: I would name this variable `concurrency`, as it is used in the semaphore.

pkg/check/uploadstress/uploadstress.go
```diff
@@ -0,0 +1,108 @@
...
+	nodeCount := int(math.Round(float64(len(clients)*o.UploadNodesPercentage) / 100))
+	rnd := random.PseudoGenerator(o.Seed)
+	picked, _ := randomPick(rnd, nodeNames, nodeCount)
+
+	rnds := random.PseudoGenerators(rnd.Int63(), nodeCount)
+
+	uGroup := new(errgroup.Group)
+	uSemaphore := make(chan struct{}, buffer)
+	for i, p := range picked {
+		o := o
```
janos, line 51: it is not needed to reassign `o` as it is always the same.

pkg/check/uploadstress/uploadstress.go
```diff
@@ -0,0 +1,108 @@
...
+			for j := 0; j < o.FilesPerNode; j++ {
+				file := bee.NewRandomFile(rnds[i], "filename", o.FileSize)
+
+				retryCount := 0
+				for {
+					retryCount++
+					if retryCount > o.Retries {
+						return fmt.Errorf("file %s upload to node %s exceeded number of retires", file.Address().String(), overlay)
+					}
+					time.Sleep(o.RetryDelay)
```
janos, line 76: Break sleep on context done, even the default retry is only 1s, but can be set to longer.
```suggestion
select {
case <-time.After(o.RetryDelay):
case <-ctx.Done():
    return ctx.Err()
```

janos (approved):
Good for the first variant. Next stress tests can have more features in terms of bringing it closer to the scenarios in real world. I think that some variants of the test should be run with more nodes creating different topologies and also to validate the uploaded data. Or at least to simulate that retrievals are happening.

pkg/check/uploadstress/uploadstress.go
svetomir: Done.

pkg/check/uploadstress/uploadstress.go
svetomir: Done.

pkg/check/uploadstress/uploadstress.go
svetomir: Done.

janos (approved):
Thanks.

Eknir:
@svetomir , could you talk me through what is happening here in our next meeting?

svetomir:
> @svetomir , could you talk me through what is happening here in our next meeting?

sure

pkg/check/uploadstress/uploadstress.go
```diff
@@ -0,0 +1,111 @@
...
+						return fmt.Errorf("file %s upload to node %s exceeded number of retires", file.Address().String(), overlay)
+					}
+					select {
+					case <-time.After(o.RetryDelay):
+					case <-ctx.Done():
+						return ctx.Err()
+					}
+
+					if err := n.UploadFile(ctx, &file, false); err != nil {
+						fmt.Printf("uploading file %s to node %s: %v\n", file.Address().String(), overlay, err)
```
acud, line 82: would be nice to have something in this string that indicates that an error happened

pkg/check/uploadstress/uploadstress.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := uGroup.Wait(); err != nil {
+		return err
+	}
+
+	fmt.Println("upload stress check completed successfully")
+	return
+}
+
+// randomPick randomly picks n elements from the list, and returns lists of picked and unpicked elements
+func randomPick(rnd *rand.Rand, list []string, n int) (picked, unpicked []string) {
```
acud, line 104: not sure if unpicked in necessary, as it is not used anywhere. also, when running this with a high number of picks, you can be certain that all nodes will be picked, especially as the list shrinks in every iteration

acud (approved):
minor changes, other LGTM. It would be nice to tell this test to run for a certain amount of _time_ rather then a specified number of iterations, which can take a varying degree of time to execute depending on different parameters of the cluster. So for example it would be nice to `./beekeeper check uploadstress --timeout 2h`.

Also, consider taking this test out of the `check` folder into some different category, as it is quite different in nature to the other checks with which it shares the folder...

svetomir:
@acud I pushed new commit resolving all of you comments.

acud (approved):
Don't stress! :)

PR #116, opened 2021-03-24
Add flag whether to expect settlements to happen during test run

metacertain:


pkg/check/settlements/settlements.go
```diff
@@ -75,7 +76,7 @@ func Check(c *bee.Cluster, o Options, pusher *push.Pusher, pushMetrics bool) (er
 			if err != nil {
 				return err
 			}
-			settlementsHaveHappened(settlements, previousSettlements)
+			settlementsHaveHappened(settlements, previousSettlements, o.ExpectSettlements)
```
ralph-pichler, line 79: rather than passing this, put the entire call into an if

acud (approved):

ralph-pichler (approved):

PR #119, opened 2021-03-29
add a test for cashouts

ralph-pichler:
This adds a check to beekeeper to test cashout functionality which currently has no coverage in tests using an actual blockchain.

It cashes out all cheques with uncashed amounts and waits for the receipt to be available and verifies that the total balance of the node which cashes out does actually change (this is not guaranteed if the other node also has a cheque in the other direction with exactly the same amount but this should be incredibly unlikely). 
All cashout transactions happen in parallel (thus this is also testing the nonce manager) an it waits at most 50s for all of them to confirm (which should always be enough for the testing blockchain). If there are no cheques to cashout, the check simply passes without doing anything.


svetomir (approved):
LGTM

acud (approved):

PR #122, opened 2021-04-04
pss timeout and address prefix option

esadakar:


cmd/beekeeper/cmd/check_pss.go
```diff
@@ -55,15 +58,19 @@ sends PSS messages to random nodes to check if WebSocket connections receive the
...
+				RequestTimeout: c.config.GetDuration(optionTimeout),
+				AddressPrefix:  c.config.GetInt(optionAddrPrefix),
 			}, pusher, c.config.GetBool(optionNamePushMetrics))
 		},
 		PreRunE: c.checkPreRunE,
 	}
 
 	cmd.Flags().Int64P(optionNameSeed, "s", 0, "seed for choosing random nodes; if not set, will be random")
+	cmd.Flags().Duration(optionTimeout, time.Minute*5, "timeout duration for pss retrieval")
+	cmd.Flags().Int(optionAddrPrefix, 2, "public address prefix bytes count")
```
acud, line 73: since this is hex, then 2 characters are needed to describe 1 byte, so the default value here should be `1`, and the number of bytes to slice from the overlay address should be `2*prefix`

acud (approved):

PR #123, opened 2021-04-06
pushsync replication loop fix

esadakar:


esadakar:
pushsync beekeeper fix

svetomir (approved):

pkg/check/pushsync/check_chunks.go
```diff
@@ -53,22 +55,26 @@ func CheckChunks(c *bee.Cluster, o Options) (err error) {
...
+				name, address, err := chunk.ClosestNodeFromMap(overlays, skipPeers...)
+				skipPeers = append(skipPeers, address)
+				if err != nil {
+					continue
+				}
+				synced, err = ng.NodeClient(name).HasChunk(ctx, ref)
+				if err != nil {
+					continue
+				}
+				if synced {
```
acud, line 70: in a small cluster this can easily end up as the actual uploader node. can we put the uploader overlay in the skip peers?

acud (changes requested):

pkg/check/pushsync/check_chunks.go
esadakar: ok

esadakar: then, as an observation, with this requirements, the cluster size cannot ever been less than three

acud (approved):

PR #134, opened 2021-04-22
Update bee client pinning API

mrekucci:


esadakar (approved):

esadakar:
Will merge once we get a 🟢 run 👍

pkg/check/gc/reserve.go
```diff
@@ -70,32 +69,38 @@ func CheckReserve(c *bee.Cluster, o Options) error {
 	}
 	fmt.Println("reservestate:", state)
 
-	lowValueChunks := chunkBatch(rnd, overlay, int64(o.CacheSize), origState.Radius)
+	pinnedChunk := bee.GenerateRandomChunkAt(rnd, overlay, 0)
+	_, err = client.UploadChunk(ctx, pinnedChunk.Data(), api.UploadOptions{Pin: true, BatchID: batchID})
+	if err != nil {
+		return fmt.Errorf("unable to upload chunk: %w", err)
+	}
+
```
acud, line 77: @mrekucci can we add a `Printf` here with the pinned chunk address? similar to what we have in `L85` - `uploaded pinned chunk %s`...

pkg/check/gc/reserve.go
```diff
@@ -140,19 +145,19 @@ func CheckReserve(c *bee.Cluster, o Options) error {
 	}
 
 	// STEP 3: Upload chunks with high value batch, then create a low value batch, and confirm no chunks were garbage collected
```
acud, line 147: before this step, I would add an explicit check that the node `Has(pinnedChunk.Address()`

pkg/check/gc/reserve.go
```diff
@@ -178,6 +183,43 @@ func CheckReserve(c *bee.Cluster, o Options) error {
...
+		return fmt.Errorf("address mismatch: have %q; want none", have)
+	}
+
+	pinned, err = client.GetPins(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to get pins: %w", err)
+	}
+	if len(pinned) > 0 {
+		return errors.New("pin count is greater than zero")
+	}
```
acud, line 221: r: a potential future improvement would be to check that the chunk is actually GCd after you unpin it, but it might need a bit more changes in the test scenario to accommodate for that. this is good as a first iteration :+1:

acud (commented):
Looks good, I have a few minor comments

pkg/check/gc/reserve.go
mrekucci: Done

pkg/check/gc/reserve.go
mrekucci: Done

pkg/check/gc/reserve.go
mrekucci: Agree with this, In the next iteration, I would even consider factoring out the local pinning test to its own separate check.

pkg/check/gc/reserve.go
acud: I meant to add a check that the node indeed stores the chunk and that it has not been GCd:
```go
has, err := client.HasChunk(ctx, pinnedChunk.Address())
.....
if !has { return error("expected node pin uploaded chunk") }
```

acud (approved):
minor addition, otherwise LGTM

pkg/check/gc/reserve.go
mrekucci: Sorry, my bad, fixed.

acud (approved):
thank you :+1:

esadakar:
```
bee nodes added
gc: reserve check
Seed: 6949947103656113963
node bee-1
reservestate: {5 1024 0 0}
created batch id 6215ff34cbe856bb552327db8ebccfa00f69adb37553e8784dfdcf8e8ba503e7 with depth 16 and amount 3
reservestate: {5 0 2 1}
uploaded pinned chunk "07f7806f27850fdd2d01683a44b1d11aed49b5bc239dbea22b469a81a41a3eff"
uploaded 100 chunks with batch depth 16, amount 1, at radius 5
uploaded 10 chunks with batch depth 16, amount 1, at radius 6

created batch id 9b126d95c7bcddc7738fad52be636ebb01990e025e5a7c3b7edbc77e018cba0c with depth 16 and amount 3
reservestate: {6 0 4 1}
retrieved low value chunks: 90, gc'd count: 10
retrieved low value high radius chunks: 10, gc'd count: 0
uploaded 100 chunks with batch depth 16, amount 3, at radius 6
created batch id 179705f8cbfae96905f7809c67d2c703e7227516862c7315938d9cb32b03ee2d with depth 16 and amount 3
reservestate: {7 128 1 1}
retrieved high value chunks: 100, gc'd count: 0
```

PR #137, opened 2021-04-26
gc: extra test case

esadakar:


esadakar:
gc two cases

pkg/check/gc/reserve.go
```diff
@@ -27,68 +38,135 @@ func CheckReserve(c *bee.Cluster, o Options) error {
...
+		if has, _ := client.HasChunk(ctx, c.Address()); has {
+			hasCount++
 		}
 	}
 
-	fmt.Printf("retrieved low value chunks: %d, gc'd count: %d\n", hasCounter, len(lowValueChunks)-hasCounter)
+	fmt.Printf("retrieved low value chunks: %d, gc'd count: %d\n", hasCount, len(lowValueChunks)-hasCount)
 
-	if len(lowValueChunks) == hasCounter {
+	if len(lowValueChunks) == hasCount {
```
acud, line 118: let's have a concrete check here that it _has_ to equal to the amount of chunks we expect to be GCd

pkg/check/gc/reserve.go
```diff
@@ -27,66 +38,134 @@ func CheckReserve(c *bee.Cluster, o Options) error {
 	}
 	fmt.Printf("node %s\n", node.Name())
 
-	adrs, err := node.Client().Addresses(ctx)
+	const (
+		lowValue  int64 = 1
```
mrekucci, line 42: r: since they represent amount I'd rename them to `lowAmonut` and `highAmount`

mrekucci (commented):
Is there a reason why the error returned in the following construct: `has, _ := client.HasChunk(ctx, c.Address())` used several times in the pkg/check/gc/reserve.go is ignored?

cmd/beekeeper/cmd/check_gc.go
```diff
@@ -138,37 +137,20 @@ func (c *command) initCheckGc() *cobra.Command {
...
-			return gc.CheckChunkNotFound(cluster, gc.Options{
-				NodeGroup:        "bee",
-				StoreSize:        c.config.GetInt(optionNameDbCapacity),
-				StoreSizeDivisor: c.config.GetInt(optionNameDivisor),
-				Wait:             c.config.GetDuration(optionNameWaitBeforeCheck),
-				Seed:             seed,
-				PostageAmount:    c.config.GetInt64(optionNamePostageAmount),
-				PostageWait:      c.config.GetDuration(optionNamePostageBatchhWait),
+			return gc.CheckReserve(cluster, gc.Options{
+				NodeGroup:     "nodes",
```
anatollupacescu, line 141: is this being used anywhere?

acud (approved):

PR #148, opened 2021-05-17
fix: new bigint serialization

AuHau:
This PR tackles the Bee BigInt serialization PR (https://github.com/ethersphere/bee-js/pull/290).

It compiles but I did not manage to actually run it as I am running on macOS and bee-local has problems with macOS :-( Could somebody please verify it functions as expected?

Also I have duplicated the `bigint` package from Bee as I don't really now how to include it from package as in the Bee it is not yet merged into master, but in order for it to be merged Beekeeper's tests needs to be working. Once it is merged then I can rework it to use the package directly from Bee.

### DIFF
--- a/pkg/bee/client.go
+++ b/pkg/bee/client.go
@@ -106,7 +106,7 @@ func (c *Client) Balance(ctx context.Context, a swarm.Address) (resp Balance, er
 	}
 
 	return Balance{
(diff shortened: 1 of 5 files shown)


svetomir:
Closed due to https://github.com/ethersphere/beekeeper/pull/150

PR #150, opened 2021-06-02
Big int serialization

svetomir:
This PR is newer version of https://github.com/ethersphere/beekeeper/pull/148.

Description is written by @AuHau and copied from https://github.com/ethersphere/beekeeper/pull/148:

This PR tackles the Bee BigInt serialization PR (ethersphere/bee-js#290).

It compiles but I did not manage to actually run it as I am running on macOS and bee-local has problems with macOS :-( Could somebody please verify it functions as expected?

Also I have duplicated the bigint package from Bee as I don't really now how to include it from package as in the Bee it is not yet merged into master, but in order for it to be merged Beekeeper's tests needs to be working. Once it is merged then I can rework it to use the package directly from Bee.

### DIFF
--- a/pkg/bee/client.go
+++ b/pkg/bee/client.go
@@ -106,7 +106,7 @@ func (c *Client) Balance(ctx context.Context, a swarm.Address) (resp Balance, er
 	}
 
 	return Balance{
-		Balance: b.Balance,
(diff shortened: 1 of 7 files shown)


acud (approved):
LGTM, thanks!

PR #154, opened 2021-06-09
retrieval simulation

janos:
This PR creates a chunk retrieval simulation, similar to the retrieval check where chunks are uploaded to nodes and downloaded to validate that they are retrievable.

The simulation is in form of a continuous set of uploads and downloads with configurable number of nodes, chunks per node and delay between uploads.

The result of the simulation are metrics beekeeper_simulation_retrieval_*.

There is no defined time for which the simulation is supposed to run. It runs until the context is canceled. In practice, until the beekeeper process is terminated.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -331,6 +331,17 @@ simulations:
       upload-node-percentage: 50
     timeout: 5m
     type: upload
+  retrieval:
+    options:
+      chunks-per-node: 1
+      metrics-enabled: true
+      postage-amount: 1
+      postage-depth: 16
+      postage-wait: 5s
+      upload-node-count: 1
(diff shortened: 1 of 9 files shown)


config/config.yaml
```diff
@@ -331,6 +331,17 @@ simulations:
...
+    options:
+      chunks-per-node: 1
+      metrics-enabled: true
+      postage-amount: 1
+      postage-depth: 16
+      postage-wait: 5s
+      upload-node-count: 1
+      upload-delay: 10s
+    timeout: 5m
+    type: retireval
```
svetomir, line 344: typo - retrieval

svetomir (approved):

PR #167, opened 2021-06-18
fix: gc chunks number

acud:
this change is needed due to the upper bound check of the eviction target which was changed recently


pkg/check/gc/reserve.go
```diff
@@ -135,7 +135,7 @@ func (c *Check) Run(ctx context.Context, cluster *bee.Cluster, opts interface{})
 
 	// upload half the CacheSize again so that reserve eviction kicks in again
```
esadakar, line 136: Please update the comment to reflect the change.

esadakar (approved):

PR #177, opened 2021-07-20
feat: propagate message from bee http error responses

janos:
This PR changes how HTTP error responses are handled in `api` and `debugapi` packages. Bee sends a json-encoded body with `"message"` field which may contain some information about the error. It is sent for all responses, not just for the bad request and no `"errors"` field is sent. This PR adjusts the error handling to include additional information in the error message, to unify errors by their code, but does not deduplicate implementations into a single package, just adjusts similar code in both packages, as that should be a part of a different PR.

### DIFF
--- a/pkg/beeclient/api/api.go
+++ b/pkg/beeclient/api/api.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
+	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


acud (approved):
Thanks @janos, much needed change!

mrekucci (approved):

svetomir (approved):

PR #178, opened 2021-07-26
feat: add content availability check

mrekucci:


notanatol (approved):

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	fmt.Printf("%s: %s\n", node, overlays[node])
+
+	batchID, err := client.GetOrCreateBatch(ctx, o.PostageAmount, o.PostageDepth, o.GasPrice, o.PostageLabel)
+	if err != nil {
+		return fmt.Errorf("node %s: unable to create batch id: %w", node, err)
+	}
+	fmt.Printf("node %s: batch id %s\n", node, batchID)
+
```
istae, line 89: You may need to sleep a bit here for the network to get notified of the batch.

istae (approved):

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+	time.Sleep(5 * time.Second) // Wait for nodes to get notified of the batch..
```
acud, line 90: this is not needed, since the GetOrCreateBatch will already wait for the batch to be `usable`

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+	isRetrievable, err := client.IsRetrievable(ctx, contentAddr)
+	if err != nil {
+		return fmt.Errorf("node %s: unable to check if content is retrievable: %w", node, err)
+	}
+	if !isRetrievable {
+		return fmt.Errorf("node %s: the uploaded content is not retrievable", node)
+	}
+	fmt.Printf("node %s: uploaded content is retrievable\n", node)
+
+	addresses := store.Addresses()
```
acud, line 109: this check can be done after the pipeline has finished crunching the content, otherwise executing the stuff in between doesn't make much sense

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+
+	rmCount := 0
+	rmChAddr := addresses[len(addresses)-1]
+	for node, nClient := range clients {
+		if err := nClient.RemoveChunk(ctx, rmChAddr); err != nil {
+			return fmt.Errorf("node %s: unable to remove chunk %s: %w", node, rmChAddr, err)
+		}
+		fmt.Printf("node %s: chunk %s removed\n", node, rmChAddr)
+
+		if rmCount++; rmCount == len(clients) {
```
acud, line 122: isn't the `for range` loop supposed to take care of this?

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+			break
+		}
+
+		time.Sleep(5 * time.Second) // Wait for nodes to sync.
```
acud, line 126: there's nothing to sync

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+
+		isRetrievable, err = client.IsRetrievable(ctx, contentAddr)
```
acud, line 128: i would execute this once after the for loop is over. this check inside the for loop could flake due to changes in connectivity patterns

pkg/check/contentavailability/contentavailability.go
```diff
@@ -0,0 +1,178 @@
...
+			exists[i] = true
+			continue
+		}
+		pm.chunks = append(pm.chunks, ch)
+		pm.exists[key] = ch.Address()
+	}
+	return exists, nil
+}
+
+func (pm *putterMock) Chunks() []swarm.Chunk {
```
acud, line 168: can't find where this is used. I think that since you don't really care about the actual chunk data once `Put` is called, you could get away with just having a map of `map[string]struct{}` that would just hold the chunk addresses

acud (commented):
Very nice. I have a few suggestions

pkg/check/contentavailability/contentavailability.go
mrekucci: Done.

pkg/check/contentavailability/contentavailability.go
mrekucci: We don't want to execute the positive check for the last node when the chunk is removed. after that the content shouldn't be retrievable.

pkg/check/contentavailability/contentavailability.go
mrekucci: Removed.

pkg/check/contentavailability/contentavailability.go
mrekucci: Removed.

pkg/check/contentavailability/contentavailability.go
mrekucci: Done.

PR #188, opened 2021-08-18
feat: authenticated check

notanatol:


pkg/beeclient/debugapi/health.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+	"net/http"
+)
+
+// HealthService represents Bee's health service
+type HealthService service
+
+// Check fetches health status of the instance
+func (b *HealthService) Check(ctx context.Context, securityToken string) (resp HealthResponse, err error) {
```
svetomir, line 12: I would rename this method to Status. Check can be misguiding.

svetomir (approved):

pkg/check/auth/auth.go
```diff
@@ -58,6 +58,9 @@ func (c *Check) Run(ctx context.Context, cluster *bee.Cluster, opts interface{})
 	for _, node := range restricted.Nodes() {
 		client := node.Client()
 
+		fmt.Println("trying to authorize with", o.Role, o.AdminUsername, o.AdminPassword)
```
mrekucci, line 61: You should replace "authorize" with "authenticate" since we're doing authentication. Also on line 66 below.

mrekucci (approved):

PR #197, opened 2021-09-19
feat: metrics on by-default

acud:
This PR simplifies the development flow by removing the need for turning on the metrics in the individual checks and instead upgrades metrics to be first-class-citizen in beekeeper `Action` interface by simply exposing the `metrics.Reporter` interface in any check that would like to expose metrics. The global defaults for collecting metrics have been changed to be on-by-default.

This eliminates any coupling inside the checks that would previously register the collectors and modify the pusher formatting mode (which is always the same). Periodic and last pushing of metrics (which was also manually done in checks, and was also error prone) is now handled in the abstraction level (inside the `cmd` package).


pkg/check/pushsync/pushsync.go
```diff
@@ -73,25 +70,16 @@ func (c *Check) Run(ctx context.Context, cluster *bee.Cluster, opts interface{})
...
 		return checkLightChunks(ctx, cluster, o)
 	default:
-		return defaultCheck(ctx, cluster, o)
+		return c.defaultCheck(ctx, cluster, o)
 	}
 }
 
 // defaultCheck uploads given chunks on cluster and checks pushsync ability of the cluster
-func defaultCheck(ctx context.Context, c *bee.Cluster, o Options) error {
+func (chk *Check) defaultCheck(ctx context.Context, c *bee.Cluster, o Options) error {
```
mrekucci, line 78: The receiver name should be consistent `chk` -> `c`. Maybe rename `bee.Cluster` `c` -> `bc`.

pkg/check/pingpong/metrics.go
```diff
@@ -1,24 +1,40 @@
...
+
+func newMetrics() metrics {
+	subsystem := "check_pingpong"
+	return metrics{
+		RttGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: mm.Namespace,
+				Subsystem: subsystem,
+				Name:      "rtt_duration_seconds",
+				Help:      "Ping round-trip time duration Gauge",
```
mrekucci, line 22: To stay consistent add the missing period (`.`) at the end of the sentence, also below.

mrekucci:
Very nice!

mrekucci (approved):

svetomir (approved):

svetomir:
Good work, trully valuable improvement.

pkg/simulate/retrieval/retrieval.go
```diff
@@ -47,11 +44,13 @@ func NewDefaultOptions() Options {
...
 // Simulation instance
-type Simulation struct{}
+type Simulation struct {
+	metrics metrics
+}
 
 // NewSimulation returns new upload simulation
 func NewSimulation() beekeeper.Action {
-	return &Simulation{}
+	return &Simulation{newMetrics("")}
```
janos, line 53: As all metrics are collected to the single database and the pusher is used which makes no initial values push, the previous and next runs are connected in the graph as the last value from the previous run is actually the first one from the next one. This is misleading and separating runs with label helps identifying the correct values for a particular simulation run.

This can be reintroduced later, again.

janos (approved):
Nice. I have only one comment, explanation on the runID, which can be addressed later if needed.

pkg/simulate/retrieval/retrieval.go
acud: :+1: the reason I left it the way it is is that I did not want to pollute the `NewSimulation` ctor with data like the cluster name. I think that once this is necessary we could find some other way to create the differentiation, maybe with some just the unix epoch when the job was started, or something along those lines, since the previous ctor had the cluster name in it

PR #214, opened 2021-11-11
fix: http request body close

istae:


pkg/bee/api/files.go
```diff
@@ -14,7 +14,7 @@ import (
 type FilesService service
 
 // Download downloads data from the node
-func (f *FilesService) Download(ctx context.Context, a swarm.Address) (resp io.ReadCloser, err error) {
+func (f *FilesService) Download(ctx context.Context, a swarm.Address) ([]byte, error) {
```
janos, line 17: This can potentially lead to resource exhaustion with larger files. The ReadCloser is deliberately returned to signal that it has to be closed.

pkg/bee/api/api.go
```diff
@@ -237,21 +239,6 @@ func (c *Client) requestWithHeader(ctx context.Context, method, path string, hea
...
-func drain(r io.ReadCloser) {
-	go func() {
-		// Panicking here does not put data in
-		// an inconsistent state.
-		defer func() {
-			_ = recover()
-		}()
-
-		_, _ = io.Copy(ioutil.Discard, r)
-		r.Close()
```
janos, line 251: This is where http body is closed. The drain function additionally reads all remaining body as the connection is not closed as the body is not read completely.

pkg/bee/api/files.go
istae: The bug I had to dig deep to find was that the reader was not closed where this  method was used. We are not at the level of testing with large enough files that would lead to resource exhaustion, and returning bytes is more explicit.

 I don't mind adding the missing Close statement instead, either way works.

pkg/bee/api/api.go
istae: `Close()` is normally enough, no draining is needed. The issue the PR tries to solve is that there were missing Close statements.

pkg/bee/api/files.go
istae: I liked the explicit return of bytes cause the current implementation has to led a bugs in a couple of places.
Not sure if we are at the position of testing with files larger than a few MB.

pkg/bee/api/api.go
istae: Normally the `Close()` is enough, no need to drain. The issue at hand is that the Close is not called in a few places.

janos: Actually, there are reports that draining is needed in order to reuse connection. Some sources: [here](https://forum.golangbridge.org/t/do-i-need-to-read-the-body-before-close-it/5594/11), [here](https://github.com/golang/go/issues/15703) and [here](https://go-review.googlesource.com/c/go/+/21290/). There is a possibility to create TIME_WAIT connections if the response body is not read. This prevents that possibility.

pkg/bee/api/files.go
janos: I see this as a regression to put the whole body in memory instead just to add a Close or drain to the returned io.ReadCloser.

Assuming the size of the body makes the implementation more fragile for the people in the future.

acud (approved):

janos (approved):

PR #229, opened 2021-12-20
Support reading configuration from a Git repo

svetomir:
This PR adds support for reading Beekeeper's configuration from a Git repo.
Reading configuration from a local machine is kept as a default behaviour.

In case configuration should be read from a Git repo, fields: _config-git-repo_ and _config-git-branch_ should be set in the "~/.beekeeper.yaml" file. Fields _config-git-username_ and _config-git-password_ can be set for a private repo.

If field _config-git-repo_ is set, it will override reading configuration from a local machine.

Official repo with Beekeeper's configuration is: https://github.com/ethersphere/beekeeper-config


vandot (approved):

janos (approved):
Wonderful!

PR #245, opened 2022-04-04
fix: print uploaded file hash

notanatol:


pkg/check/smoke/smoke.go
```diff
@@ -88,104 +88,131 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
+			}()
 		}
 
-		// We want to update the metrics when no error has been
-		// encountered in order to avoid counter mismatch.
-		c.metrics.UploadDuration.Observe(txDuration.Seconds())
-		c.metrics.DownloadDuration.Observe(rxDuration.Seconds())
+		for r := range resp {
+			if r.err != nil {
+				fmt.Println(err)
```
mrekucci, line 213: Should be `r.err` not `err`.

pkg/check/smoke/smoke.go
```diff
@@ -88,104 +88,131 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
+
+				resp <- downloadRes{}
+			}()
 		}
 
-		// We want to update the metrics when no error has been
-		// encountered in order to avoid counter mismatch.
-		c.metrics.UploadDuration.Observe(txDuration.Seconds())
-		c.metrics.DownloadDuration.Observe(rxDuration.Seconds())
+		for r := range resp {
```
mrekucci, line 211: This will give you a deadlock! Use `sync.WaitGroup`.

mrekucci (changes requested):
When deploying concurrent download clients, log messages that do not contain the client name should be corrected:
```
fmt.Printf("upload failed: %v\n", err)
fmt.Printf("retrying in: %v\n", o.TxOnErrWait)
...
fmt.Printf("retrying in: %v\n", o.RxOnErrWait)
...
fmt.Errorf("length mismatch: rx length %d; tx length %d", rxLen, txLen),
...
fmt.Errorf("data mismatch:found %d different bytes", diff)
```

aloknerurkar:
@notanatol I dont think these changes are needed for now. I think we can always add more tests for parallel upload/download cases but then it can be a separate test in the check pkg.

PR #257, opened 2022-05-17
test: add unit testing for K8s package

gacevicljubisa:
Write unit tests for package k8s.


pkg/random/random_test.go
```diff
@@ -99,16 +99,48 @@ func TestPseudoGenerators(t *testing.T) {
 	}
 }
```
mrekucci, line 101: Such tests do not have much added value. Instead of covering one random value, you should cover a range or better option would be to use fuzzy testing: https://go.dev/doc/fuzz/

gacevicljubisa: Thanks for the link in the comment, I will investigate and fix it.

pkg/k8s/containers/security.go
```diff
@@ -76,3 +81,11 @@ type WindowsOptions struct {
 	GMSACredentialSpec     string
 	RunAsUserName          string
 }
+
+func (wo *WindowsOptions) toK8S() *v1.WindowsSecurityContextOptions {
```
svetomir, line 85: should return nil if nothing is configured

pkg/k8s/pod/security.go
```diff
@@ -85,3 +86,11 @@ type WindowsOptions struct {
 	GMSACredentialSpec     string
 	RunAsUserName          string
 }
+
+func (wo *WindowsOptions) toK8S() *v1.WindowsSecurityContextOptions {
```
svetomir, line 90: should return nil if nothing is configured

pkg/k8s/containers/security.go
gacevicljubisa: done

pkg/k8s/pod/security.go
gacevicljubisa: done

svetomir (approved):
Good work. Thank you

PR #276, opened 2022-08-03
fix: ci topup option fix

swarmHaseeb:
Topup amount increased from initial amount to allow postage-new-depth to increase.

### DIFF
--- a/config/local.yaml
+++ b/config/local.yaml
@@ -360,7 +360,7 @@ checks:
     options:
       postage-amount: 1000
       postage-depth: 17
-      postage-topup-amount: 1100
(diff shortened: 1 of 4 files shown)


config/local.yaml
```diff
@@ -360,7 +360,7 @@ checks:
     options:
       postage-amount: 1000
       postage-depth: 17
-      postage-topup-amount: 100
+      postage-topup-amount: 1100
```
istae, line 363: let's go back up 100, this should also works. If it doesn't work, we have to understand why.

istae (changes requested):

istae (approved):

PR #284, opened 2022-11-03
fix: new libp2p keys

notanatol:


vandot:
@notanatol why we need new keys? I ask because we have them in various places outside beekeeper as well.

notanatol:
> @notanatol why we need new keys? I ask because we have them in various places outside beekeeper as well.

yes, after we merge the libp2p library upgrade into master it's going to need new keys.

vandot (approved):

PR #291, opened 2022-11-23
fix: optional force new batch

notanatol:


pkg/check/smoke/load.go
```diff
@@ -109,7 +112,20 @@ func (c *LoadCheck) Run(ctx context.Context, cluster orchestration.Cluster, opts
 					c.metrics.UploadAttempts.Inc()
 					var duration time.Duration
 					c.logger.Infof("uploading to: %s", txName)
-					address, duration, err = test.upload(txName, txData)
+
+					var batchID string
+					if time.Since(start) > o.MaxUseBatch || batchID == "" { // force buy new batch
+						batchID, err = clients[txName].CreatePostageBatch(ctx, o.PostageAmount, o.PostageDepth, o.GasPrice, "load-test", true)
```
istae, line 118: the batchID needs to be saved per txName

istae (changes requested):

istae (approved):

PR #296, opened 2022-12-06
fix: add missing contract addresses

notanatol:
add missing addresses and flag


mrekucci (approved):

gacevicljubisa:
These options should be added to config yaml files as well.

PR #323, opened 2023-02-15
fix: fix bee config options

vandot:
Remove obsolete bee config options. Bee will now fail if unknown options are inside config https://github.com/ethersphere/bee/pull/3782

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -117,12 +117,10 @@ bee-configs:
     debug-api-addr: ":1635"
     debug-api-enable: true
     full-node: true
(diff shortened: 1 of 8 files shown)


vandot:
I just removed all obsolete options from the code but `transaction` was used inside beekeeper' code.
Is this correct https://github.com/ethersphere/beekeeper/pull/323/files#diff-e1d9b11e1aaae66a290d65fedd570b47ea7c3e134e366a5040efa14d877692f7L722 ?

umerm-work (approved):

PR #330, opened 2023-02-28
fix: setup cluster

gacevicljubisa:
- trigger node groups in parallel


gacevicljubisa:
moved to new branch: k8s-cluster-setup

PR #333, opened 2023-03-11
enhancement: k8s cluster setup

gacevicljubisa:
-beekeeper update to go v1.20
-add semaphore to limit number of concurrent requests to k8s API server
-add rate limiter for QPS and Burst to k8s client
-use Ethereum address if provided in swarm key for funding
-run node groups concurrently when creating cluster


svetomir (approved):
Great job

PR #359, opened 2023-10-03
feat: check long availability

acha-bill:
Adds a check for long availability. It receives a list of references and tries to download them.

### DIFF
--- a/pkg/check/longavailability/longavailability.go
+++ b/pkg/check/longavailability/longavailability.go
@@ -26,7 +26,7 @@ func NewDefaultOptions() Options {
 		RndSeed:      time.Now().UnixNano(),
 		RetryCount:   3,
 		RetryWait:    10 * time.Second,
-		NextIterWait: 30 * time.Minute,
+		NextIterWait: 6 * time.Hour,
 	}
 }
 
@@ -62,7 +62,9 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, o interf
 		addresses = append(addresses, addr)
 	}
 
-	for it := 0; true; it++ {
+	var it int
+	for {
+		it++
 		select {
 		case <-ctx.Done():
 			return nil
@@ -77,8 +79,12 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, o interf
 				continue
 			}
 
-			var i int
-			for i = 0; i < opts.RetryCount; i++ {
+			for i := 0; i <= opts.RetryCount; i++ {
+				if i == opts.RetryCount {
+					c.logger.Errorf("node %s: download for %s failed after %d tries", node.Name(), addr, opts.RetryCount)
+					break
+				}
+
 				c.metrics.DownloadAttempts.Inc()
 				c.logger.Infof("node %s: download attempt %d for %s", node.Name(), i+1, addr)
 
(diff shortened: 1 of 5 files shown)


pkg/check/longavailability/longavailability.go
```diff
@@ -0,0 +1,132 @@
...
+	downloadTimeout time.Duration
+	retryWait       time.Duration
+}
+
+func (t *test) run(ctx context.Context, addr swarm.Address) error {
+	ctx, cancel := context.WithTimeout(ctx, t.downloadTimeout)
+	defer cancel()
+
+	for i := 0; i < t.retryCount; i++ {
+		t.metrics.DownloadAttempts.WithLabelValues(addr.String()).Inc()
```
istae, line 107: labeled metrics can be tricky to deal with in grafana, so I would go for the basic counter metric.

pkg/check/longavailability/longavailability.go
```diff
@@ -0,0 +1,132 @@
...
+	client          *bee.Client
+	logger          logging.Logger
+	metrics         metrics
+	retryCount      int
+	downloadTimeout time.Duration
+	retryWait       time.Duration
+}
+
+func (t *test) run(ctx context.Context, addr swarm.Address) error {
+	ctx, cancel := context.WithTimeout(ctx, t.downloadTimeout)
```
istae, line 103: the timeout of 5 minutes may not be enough for something like 10 GB file. we probably don't want a timeout

pkg/check/longavailability/longavailability.go
```diff
@@ -0,0 +1,132 @@
...
+			client:          client,
+			logger:          c.logger,
+			metrics:         c.metrics,
+			retryCount:      opts.RetryCount,
+			downloadTimeout: opts.DownloadTimeout,
+			retryWait:       opts.RetryWait,
+		}
+		err = t.run(ctx, addr)
+		if err != nil {
+			c.logger.Infof("node %s: download for %s failed: %v", node.Name(), addr, err)
```
istae, line 84: maybe use `Errorf`

pkg/check/longavailability/longavailability.go
```diff
@@ -0,0 +1,137 @@
...
+	var addresses []swarm.Address
+	for _, ref := range opts.Refs {
+		addr, err := swarm.ParseHexAddress(ref)
+		if err != nil {
+			return fmt.Errorf("parse hex address: %w", err)
+		}
+		addresses = append(addresses, addr)
+	}
+
+	for it := 0; true; it++ {
```
istae, line 65:
```suggestion
	for {
	  it++
	}
```

pkg/check/longavailability/longavailability.go
```diff
@@ -0,0 +1,137 @@
...
+
+		for _, addr := range addresses {
+			node, err := findRandomNode(ctx, addr, cluster, opts.RndSeed)
+			if err != nil {
+				c.logger.Errorf("find node %s. Skipping. %w", addr.String(), err)
+				continue
+			}
+
+			var i int
+			for i = 0; i < opts.RetryCount; i++ {
```
istae, line 81:
```suggestion
			for i := 0; i <= opts.RetryCount; i++ {
			  if i == opts.RetryCount {
			    c.logger.Errorf("node %s: download for %s failed after %d tries", node.Name(), addr, opts.RetryCount)
			     break
			  }
			}
```

istae (approved):

PR #372, opened 2023-12-18
refactor(logger): improve and add fatal level

gacevicljubisa:


pkg/k8s/pod/client.go
```diff
@@ -84,7 +84,7 @@ func (c *Client) EventsWatch(ctx context.Context, namespace string, operatorChan
 		// TODO: add this label to beekeeper and filter on it => app.kubernetes.io/name=bee
 	})
 	if err != nil {
-		return fmt.Errorf("getting pod events in namespace %s: %w", namespace, err)
+		return fmt.Errorf("getting pod events in namespace %s`: %w", namespace, err)
```
vandot, line 87: whats with this tick?

PR #374, opened 2023-12-19
feat: net availability SLA test

istae:


pkg/check/networkavailability/check.go
```diff
@@ -0,0 +1,194 @@
...
+	ret := make([]swarm.Address, 0, max)
+
+	for i := 0; i < max; i++ {
+		buf := make([]byte, 4)
+		binary.LittleEndian.PutUint32(buf, uint32(i))
+
+		var addr []byte
+
+		if bits <= 8 {
+			addr = []byte{buf[0]}
```
mrekucci, line 171: This is unnecessary, you can just create a new slice pointing to the buf, e.g.: `addr := buf[:1]` and so on...

mrekucci (approved):

PR #375, opened 2023-12-19
fix: wait for push and fix logs

acha-bill:


istae (approved):
Remove unused code instead of commenting it out.

PR #380, opened 2024-01-16
feat: Add retry mechanism for datadurability check

acha-bill:


acha-bill:
Last year we wanted to make the durability test more reliable by adding a retry. Since then, we've not had issues with durability checks. So maybe we can close this.

martinconic (approved):

PR #389, opened 2024-03-17
chore: golang v1.22

gacevicljubisa:


gacevicljubisa:
The current workflow in the bee project pulls the Go version directly from its go.mod file, which is set to Go 1.21. As this project is looking to upgrade to Go 1.22, we face a version mismatch that prevents merging this PR as is. We either need to update the Go version in bee to 1.22 to maintain compatibility, or consider adjusting the workflow here to accommodate the existing version in bee until such an update can be made.

vandot (approved):

PR #399, opened 2024-05-17
feat: remove auth (wip)

acha-bill:
Remove authentication

The following options have been removed
- restricted
- token-encryption-key
- admin-password


acha-bill:
@vandot configs need cleanup. I'm not sure which settings to change.

istae (approved):

Issue #400, opened 2024-05-27
upload worker should take in a max storage radius option


istae:
the upload worker should take in a max storage radius option so that the uploads can stop once the radius is reached.

acquire the storage radius either from the /reservestate or /status

the idea is after picking an uploader node, the storage radius should be checked (of the uploader). if the radius is above the allowed maximum value, the upload should not happen.
the worker can then sleep for 5-10 min, then repeat the same thing. 

this task applies to the smoke/load job in beekeeper and not the smoke job

the maximum allowed radius will be a new config option passed to the job

gacevicljubisa:
Released with v0.17.0 (https://github.com/ethersphere/beekeeper/releases/tag/v0.17.0)

PR #401, opened 2024-06-03
feat(load): modify load check to check if storage radius is bigger then max

gacevicljubisa:
### Description

This PR addresses [issue #400](https://github.com/ethersphere/beekeeper/issues/400). It adds a check in the upload worker to stop uploads if the storage radius exceeds a configured maximum value.

### Purpose

To ensure uploads only proceed when the storage radius is within acceptable limits.

### Implementation

- Checks storage radius from `/reservestate` or `/status` before uploading.
- If the radius is too high, the worker sleeps for 5-10 minutes before retrying.
- Applies to the smoke/load job, not the smoke job.
- Introduces a new config option for the max allowed radius.

### Testing

I need reviewers to verify the flow and suggest improvements. Do we need a new config option for the sleep time, or can we utilize an existing one? Currently, three requests to `/reservestate` are made before uploading: one that is added and two in `CreatePostageBatch` (one prior to buying the batch and one after buying the batch). Please provide the correct parameters for the load check to reproduce the behavior.


pkg/check/smoke/load.go
```diff
@@ -102,24 +112,46 @@ func (c *LoadCheck) Run(ctx context.Context, cluster orchestration.Cluster, opts
...
+						if rs.StorageRadius <= o.MaxStorageRadius {
+							break
+						}
+						c.log.Infof("waiting for StorageRadius to decrease. Current: %d, Max: %d", rs.StorageRadius, o.MaxStorageRadius)
+
+						select {
+						case <-ctx.Done():
+							c.log.Infof("context done in StorageRadius check: %v", ctx.Err())
+							return
+						case <-time.After(o.TxOnErrWait): // TODO: Adjust the sleep duration as needed
```
martinconic, line 142: What should be the adjustment here?

gacevicljubisa: Regarding context check, if global timeout happens, the context should be canceled, and we should stop execution.

Regarding the note: 
> **TODO:** Adjust the sleep duration as needed

I have already put a question in the PR description:

> Do we need a new config option for the sleep time, or can we utilize an existing one?

pkg/check/smoke/load.go
```diff
@@ -102,24 +112,46 @@ func (c *LoadCheck) Run(ctx context.Context, cluster orchestration.Cluster, opts
...
 					}
 
+					// Check the StorageRadius
+					for {
+						rs, err := clients[txName].ReserveState(ctx)
+						if err != nil {
+							c.log.Infof("error getting state: %v", err)
+							return
+						}
+						if rs.StorageRadius <= o.MaxStorageRadius {
```
istae, line 133: should be strictly `<`

pkg/check/smoke/load.go
istae: this check does not need to happen so often, something like every 5 min, at the max.
let's add a new config option for this wait, with default being 5 min.

istae (changes requested):

istae (approved):

PR #403, opened 2024-07-02
fix: toolchain not available

ferencsarai:
`make binary`
`go: download go1.22 for darwin/arm64: toolchain not available`

see: [this comment](https://github.com/golang/go/issues/62278#issuecomment-1933790368)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ethersphere/beekeeper
 
-go 1.22
+go 1.22.0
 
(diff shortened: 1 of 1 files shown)


gacevicljubisa:
Did you try to install newer version of golang @ferencsarai ?

ferencsarai:
SOLVED!

go 1.21.4: works with `go 1.22.0`
after install go 1.22.4 works with `go 1.22`

PR #411, opened 2024-08-23
refactor: extract label-selector as a flag for node-funder and operator

gacevicljubisa:
Refactor: Extract label-selector as a flag with default `beekeeper.ethswarm.org/node-funder=true`.

- Introduced a flag to specify the label selector for K8S resources used by node-funder and operator.
- Default label selector set to `beekeeper.ethswarm.org/node-funder=true`.
- If the label selector is an empty string, no filtering is applied.
- The label selector works in conjunction with the namespace for more granular resource selection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ It has following flags:
 --min-native float        Minimum amount of chain native coins (xDAI) nodes should have.
 --min-swarm float         Minimum amount of swarm tokens (xBZZ) nodes should have.
 --namespace string        Kubernetes namespace. Overrides cluster name if set.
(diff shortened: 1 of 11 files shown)


vandot (approved):

gacevicljubisa:
> This should be installed in-cluster or no ? so why ingress ? in cluster communication should go via services

It will 

> Member

@darkobas2 `node-funder` is not meant to work in-cluster only, but the `operator` feature is meant to work in-cluster. I will adapt that as well to use the same label selector as default, but with the option to explicitly set it.

PR #417, opened 2024-09-20
fix: config setup for hetzner k8s

gacevicljubisa:


istae (approved):

config/config.yaml
```diff
@@ -81,15 +67,14 @@ node-groups:
       nginx.ingress.kubernetes.io/session-cookie-name: "SWARMGATEWAY"
       nginx.ingress.kubernetes.io/session-cookie-path: "default"
       nginx.ingress.kubernetes.io/ssl-redirect: "true"
-    ingress-class: "nginx-internal"
+      nginx.ingress.kubernetes.io/whitelist-source-range: "142.132.221.245/32" #VPN IP, or your IP (but ethswarm.org is needed)
```
darkobas2, line 70: Not needed now, vpn is whitelisted globally

gacevicljubisa: Still getting the `403 Forbidden`if I do not whitelist... Does it depend on VPN config?

config/staging.yaml
```diff
@@ -31,4 +31,10 @@ node-groups:
 bee-configs:
   staging:
     _inherit: "default"
-    bootnodes: "/dnsaddr/mainnet.ethswarm.org"
+    swap-endpoint: http://rpc-sepolia-haproxy.default.svc.swarm1.local
```
gacevicljubisa, line 34: @darkobas2 can we use this endpoint in case of setting up "our" staging env that will be connected to sepolia?

darkobas2: This will work on our kubernetes clusters... Not on a local cluster or so.

config/config.yaml
darkobas2: No, its intended to resolve testnet.internal... maybe you still have a 'public' ingress

config/staging.yaml
gacevicljubisa: This setup is for our kubernetes clusters, for example to deploy it in some of our namespaces..

darkobas2: Yes

config/config.yaml
gacevicljubisa: @darkobas2 whitelist ingress anotation is removed. After the changes you made, I have tested configuration and it works without it.

darkobas2 (approved):

PR #426, opened 2024-10-22
fix: use fullnodes in manifest and postage test

acha-bill:


martinconic (approved):

pkg/check/manifest/manifest.go
```diff
@@ -91,7 +91,7 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
 		return err
 	}
 
-	sortedNodes := cluster.NodeNames()
+	sortedNodes := cluster.FullNodeNames()
```
gacevicljubisa, line 94: Are the nodes really sorted? Do they need to be?

acha-bill: I'm not sure. But the order is irrelevant

PR #433, opened 2024-11-14
feat: gsoc check

acha-bill:
Adds gsoc check. See https://github.com/anythread/gsoc/blob/master/test/index.spec.ts

Summary:
- socs uploaded sequentially to `node1` should arrive `node2` on the subscribed web socket.
- socs uploaded in parallel to `node1` should also arrive `node2` on the subscribed web socket.

### DIFF
--- a/pkg/check/gsoc/gsoc.go
+++ b/pkg/check/gsoc/gsoc.go
@@ -1,0 +1,311 @@
+package gsoc
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/ethersphere/bee/v2/pkg/cac"
+	"github.com/ethersphere/bee/v2/pkg/crypto"
+	"github.com/ethersphere/bee/v2/pkg/soc"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+	"github.com/ethersphere/beekeeper/pkg/bee"
+	"github.com/ethersphere/beekeeper/pkg/beekeeper"
+	"github.com/ethersphere/beekeeper/pkg/logging"
+	"github.com/ethersphere/beekeeper/pkg/orchestration"
+	"github.com/gorilla/websocket"
+	"golang.org/x/sync/errgroup"
+)
+
+// Options represents check options
+type Options struct {
+	PostageAmount int64
+	PostageDepth  uint64
+	PostageLabel  string
+}
+
+// NewDefaultOptions returns new default options
+func NewDefaultOptions() Options {
+	return Options{
+		PostageAmount: 1000,
+		PostageDepth:  17,
+		PostageLabel:  "test-label",
+	}
+}
+
+// compile check whether Check implements interface
+var _ beekeeper.Action = (*Check)(nil)
+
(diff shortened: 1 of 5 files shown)


pkg/check/gsoc/gsoc.go
```diff
@@ -0,0 +1,317 @@
...
+	ch, closeWebSocket, err := listenWebsocket(ctx, listenClient.Config().APIURL.Host, socAddress, logger)
+	if err != nil {
+		return fmt.Errorf("listen websocket: %w", err)
+	}
+
+	defer func() {
+		if closeWebSocket != nil {
+			closeWebSocket()
+		}
+	}()
```
gacevicljubisa, line 130: Maybe it would be cleaner if WebSocket handiling is encapsulated in struct where `Listen` and `Close` methods are exposed. `Close` checks if the WebSocket is already closed to avoid double closures. And then here just call `defer  wsListener.Close()`

pkg/check/gsoc/gsoc.go
```diff
@@ -0,0 +1,317 @@
...
+	err = run(ctx, uploadClient, listenClient, batches, c.logger, true)
+	if err != nil {
+		return err
+	}
+	c.logger.Infof("done")
+
+	return nil
+}
+
+func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client, batches []string, logger logging.Logger, parallel bool) error {
```
gacevicljubisa, line 104: Maybe this method can be refactored, some suggestions are to extract `func uploadChunk`, use two separte func for exectution mode:
```
    if parallel {
        logger.Infof("sending messages in parallel...")
        err = runInParallel(ctx, otherParams...)
    } else {
        logger.Infof("sending messages sequentially...")
        err =  runInParallel(ctx, otherParams...)
    }
```
And in both methods iterate and call uploadChunk func.

pkg/check/gsoc/gsoc.go
```diff
@@ -0,0 +1,347 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func runInParallel(ctx context.Context, client *bee.Client, numChunks int, batches []string, resourceId []byte, privKey *ecdsa.PrivateKey, logger logging.Logger) error {
+	var errG errgroup.Group
+	for i := 0; i < numChunks; i++ {
+		i := i
```
gacevicljubisa, line 200: this `i:=i` is not required since go 1.22

pkg/check/gsoc/gsoc.go
```diff
@@ -0,0 +1,347 @@
...
+		err = runInParallel(ctx, uploadClient, numChunks, batches, resourceId, privKey, logger)
+	} else {
+		err = runInSequence(ctx, uploadClient, numChunks, batches, resourceId, privKey, logger)
+	}
+	if err != nil {
+		return err
+	}
+
+	// wait for listener to receive all messages
+	time.Sleep(5 * time.Second)
```
gacevicljubisa, line 155: Does this mean that all messages are expected to be received within this timeframe?

gacevicljubisa (approved):

martinconic (approved):

pkg/check/gsoc/gsoc.go
acha-bill: yes

PR #440, opened 2024-12-11
feat: stamper

gacevicljubisa:
This PR introduces a new command group, `stamper`, to manage postage batches for Bee nodes, and enhances the existing `node-funder` functionality. It also includes fixes and improvements related to cluster setup and configuration usage.

##### **Stamper Command**

The `stamper` command group is designed to handle postage batches with the following subcommands:

1. **`create`**:  
   Creates a new postage batch for selected nodes with customizable parameters like amount, depth, and timeout.  

2. **`topup`**:  
   Tops up postage batches for selected nodes based on specified thresholds for TTL and periodic checks.  

3. **`dilute`**:  
   Dilutes existing postage batches by increasing the depth based on utilization thresholds.  

4. **`set`**:  
   Combines the functionalities of  `topup`, and `dilute` to manage postage batches comprehensively.  

Periodic checks for `topup`, `dilute`, and `set` subcommands are handled by the `scheduler` package.

##### **Node-Funder Enhancements**

- Added `--periodic-check` flag to support periodic execution of funding operations.  

#### **Fixes and Improvements**

- Enhanced cluster setup process.
- Improved configuration handling for better flexibility and usability.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,10 +1,9 @@
 package beekeeper
 
 var (
-
 	version = "0.25.0" // manually set semantic version number
 
-	commit  string     // automatically set git commit hash
+	commit string // automatically set git commit hash
 
 	// Version TODO
 	Version = func() string {


pkg/stamper/node.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+
+			requiredDuration := int64(topupDuration.Seconds()) - remainingTTL
+
+			if requiredDuration > 0 {
+				amount := (requiredDuration / blockTime) * multiplier * price
+
+				n.log.Tracef("node %s: batch %s: required duration %d, amount %d", n.name, batch.BatchID, requiredDuration, amount)
+
+				if err := n.client.Postage.TopUpPostageBatch(ctx, batch.BatchID, amount, ""); err != nil {
```
istae, line 125: i would pass the old label to the topup request

gacevicljubisa: last argument for gas price

pkg/stamper/node.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if len(batchIDs) > 0 && !contains(batchIDs, batch.BatchID) {
+		return false
+	}
+
+	return true
+}
+
+func contains(slice []string, value string) bool {
```
janos, line 188: Just a minor thing, this function can be replaced with https://pkg.go.dev/slices#Contains. I think that performance difference is not drastic.

pkg/stamper/stamper.go
```diff
@@ -0,0 +1,269 @@
...
+	log           logging.Logger
+	namespace     string
+	k8sClient     *k8s.Client
+	swapClient    swap.Client
+	beeClients    map[string]*bee.Client
+	labelSelector string
+	inCluster     bool
+}
+
+func New(cfg *ClientConfig) *client {
```
janos, line 61: As `client` is not exported, maybe it is nicer to return the `Client` interface in this function signature.

pkg/swap/http.go
```diff
@@ -78,11 +78,7 @@ func httpClientWithTransport(baseURL *url.URL, c *http.Client) *http.Client {
 
 	c.Transport = roundTripperFunc(func(r *http.Request) (resp *http.Response, err error) {
 		r.Header.Set("User-Agent", userAgent)
-		u, err := baseURL.Parse(r.URL.String())
-		if err != nil {
-			return nil, err
-		}
-		r.URL = u
+		r.URL = baseURL
```
janos, line 81: Is it ok to ignore the original r.URL and always use baseURL instead? It looks to me that url.Parse basically merges two urls.

janos (commented):
Very nice work, Ljubisa.

pkg/swap/http.go
gacevicljubisa: Now, the full URL provided in the request (including the path and query parameters) is preserved, which was not the case earlier. Also, geth-url is provided in init of the NewGethClient, so the baseURL always stays the same. This url is RPC endpoint and it doesn't change.

pkg/stamper/stamper.go
gacevicljubisa: I have exporte the Client and removed the interface.

pkg/swap/http.go
janos: Thanks, makes sense.

pkg/stamper/stamper.go
janos: Great.

janos (approved):
👍

istae (approved):

PR #441, opened 2024-12-14
feat: feeds check

acha-bill:


pkg/check/feed/feed.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts interface{}) (err error) {
+	o, ok := opts.(Options)
+	if !ok {
+		return fmt.Errorf("invalid options type")
+	}
+
+	if o.RootRef != "" {
+		c.logger.Infof("running availability check")
+		return c.availability(ctx, cluster, o)
```
istae, line 60: maybe a better name like checkAvailability?

pkg/check/feed/feed.go
```diff
@@ -0,0 +1,174 @@
...
+
+	client := clients[nodeName]
+	_, _, err = client.DownloadFile(ctx, ref, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Check) regular(ctx context.Context, cluster orchestration.Cluster, o Options) error {
```
istae, line 86: maybe rename to something like fullFeedsCheck with some comments above the function?

pkg/check/feed/feed.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		c.logger.Infof("node %s: feed updated", nodeName)
+		c.logger.Infof("soc reference: %s", socRes.Reference)
+		c.logger.Infof("wrapped reference: %s", file.Address())
+	}
+	time.Sleep(5 * time.Second)
+
+	// fetch update
+	c.logger.Infof("fetching feed update")
+	downClient := clients[nodeNames[1]]
```
istae, line 149: client selection should ideally be random...also if the slice size is less than 2, this could lead to a panic.

istae (approved):
nice work, some comments

gacevicljubisa:
The purpose of the feed check is to enable running it as a regression test. To achieve this, we need some output from the first run that can be reused in subsequent runs (e.g., after updating the Docker image of the node).

The process of generating and retrieving this output should be straightforward. The output can be displayed in the console, but we could also store it in a file for convenience.

To simplify automation, we could follow a sequence of commands like this:

```
beekeeper check --cluster-name=local-dns --checks=feed --output=example.yaml  
beekeeper restart --cluster-name=default --image=ethersphere/bee:latest  
beekeeper check --cluster-name=local-dns --checks=feed --input=example.yaml  
```

This approach can be generalized for other regression tests in the future.

@istae, @acha-bill What are your thoughts on this?

acha-bill:
This feed check is in 2 parts—the integration and regression tests. We went with an approach similar to long availability tests for the regression test. We try to retrieve a feed that already exists on public testnet

The approach you propose is neat as well since it can be run locally or in CI without the public testnet.
If we agree to proceed with this new approach, I can strip off the regression part from this PR and keep the integration.

gacevicljubisa (approved):

PR #443, opened 2025-01-15
feat: extend manifest integration test

acha-bill:
Uses feeds to test updates to a website

1. It creates a root feed manifest
2. To make an update to the feed, it uploads a tar website with some index.html and other files. Then uses this reference in the feed update.
3. It then tries to download the website via the root feed reference. It should resolve to the index.html
4. To update the website, we push another update to the feed with the new tar and then try to download it again via the root feed reference.
5. Lastly, it tries to download the other files in the website directly. e.g `b25c89a401d9f26811680476619a1eb4a4e189e614bc6161cbfd8b343214917b/assets/styles/styles.css`

### DIFF
--- a/pkg/check/manifest/manifest.go
+++ b/pkg/check/manifest/manifest.go
@@ -77,10 +77,10 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
 	upClient := clients[0]
 	downClient := clients[1]
 
-	//err = c.checkWithoutSubDirs(ctx, rnd, o, upClient, downClient)
-	//if err != nil {
-	//	return fmt.Errorf("check without subdirs: %w", err)
-	//}
+	err = c.checkWithoutSubDirs(ctx, rnd, o, upClient, downClient)
+	if err != nil {
+		return fmt.Errorf("check without subdirs: %w", err)
+	}
 
 	err = c.checkWithSubDirs(ctx, rnd, o, upClient, downClient)
 	if err != nil {
--- a/pkg/check/manifest/manifest.go
+++ b/pkg/check/manifest/manifest.go
@@ -168,7 +168,7 @@ func (c *Check) checkWithSubDirs(ctx context.Context, rnd *rand.Rand, o Options,
 	c.logger.Infof("collection uploaded: %s", tarFile.Address())
 	time.Sleep(3 * time.Second)
 
-	//push first version of website to the feed
+	// push first version of website to the feed
 	ref, err := upClient.UpdateFeedWithReference(ctx, signer, topic, 0, tarFile.Address(), api.UploadOptions{BatchID: batchID})
 	if err != nil {
 		return err
--- a/pkg/check/manifest/manifest.go
+++ b/pkg/check/manifest/manifest.go
@@ -11,6 +11,8 @@ import (
 	"math/rand"
 	"time"
 
+	"github.com/ethersphere/bee/v2/pkg/crypto"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/ethersphere/beekeeper/pkg/bee"
 	"github.com/ethersphere/beekeeper/pkg/bee/api"
 	"github.com/ethersphere/beekeeper/pkg/beekeeper"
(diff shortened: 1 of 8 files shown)


pkg/check/manifest/manifest.go
```diff
@@ -67,14 +69,34 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
-	overlays, err := cluster.FlattenOverlays(ctx)
+	clients, err := cluster.NodesClients(ctx)
 	if err != nil {
 		return err
 	}
+	upClient := clients[names[perm[0]]]
+	downClient := clients[names[perm[1]]]
+
+	err = c.checkWithoutSubDirs(ctx, rnd, o, upClient, downClient)
+	if err != nil {
```
martinconic, line 87: If err !=nil you return here and will not `checkWithSubDirs`. Is this intentional? What if `checkWithoutSubDirs` returns an error but `checkWithSubDirs` would not return any errors?

acha-bill: Yes it's intentional. If the first check fails, the test should fail immediately

gacevicljubisa:
One more improvement regarding Options:
https://github.com/ethersphere/beekeeper/blob/4d02a31cd627a9abeeed444ff0b702ca3ec1c695/pkg/check/manifest/manifest.go#L36-L46
Maybe to put `PostageLabel` default value to `manifest-check`, or something like that to distingush it from other batch labels. 
Maybe this would be nice to do for every check that uses `PostageLabel` option, to have the check name in it? What do you think @istae ?

Also, MaxPathnameLength shouldn't be less then 2, or it will panic.

pkg/orchestration/k8s/cluster.go
```diff
@@ -233,6 +233,22 @@ func (c *Cluster) FullNodeNames() (names []string) {
 	return
 }
 
+// ShuffledFullNodeClients returns a list of full node clients shuffled
+func (c *Cluster) ShuffledFullNodeClients(ctx context.Context, r *rand.Rand) ([]*bee.Client, error) {
+	cls, err := c.NodesClients(ctx)
```
gacevicljubisa, line 238: This will return all nodes, not only full nodes, so maybe check with implementation of `FullNodeNames()` method and combine it to extract only full nodes.

istae:
> One improvement regarding Options:
> 
> https://github.com/ethersphere/beekeeper/blob/4d02a31cd627a9abeeed444ff0b702ca3ec1c695/pkg/check/manifest/manifest.go#L36-L46
> 
> 
> Maybe to put `PostageLabel` default value to `manifest-check`, or something like that to distingush it from other batch labels.
> Maybe this would be nice to do for every check that uses `PostageLabel` option, to have the check name in it? What do you think @istae ?
> Also, MaxPathnameLength shouldn't be less then 2, or it will panic.

The label can be used strategically in general, because the GetCreateMutableBatch function matches the label, and it can reuse the same batch created before to save on cost and time (by not waiting for a new batch to be created which can take few minutes).

gacevicljubisa:
> The label can be used strategically in general, because the GetCreateMutableBatch function matches the label, and it can reuse the same batch created before to save on cost and time (by not waiting for a new batch to be created which can take few minutes).

Would this mean that we can resuse the same batch created with specific check? Or is it ok to use `test-label` for all of the checks?
What is your suggestion, to leave the same label for all of the checks?

pkg/check/manifest/manifest.go
```diff
@@ -67,14 +69,31 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
 
 	rnd := random.PseudoGenerator(o.Seed)
+	clients, err := cluster.ShuffledFullNodeClients(ctx, rnd)
+	if err != nil {
+		return fmt.Errorf("node clients shuffle: %w", err)
+	}
 
-	c.logger.Infof("Seed: %d", o.Seed)
+	if len(clients) < 2 {
+		return fmt.Errorf("not enough nodes to run feed check")
```
gacevicljubisa, line 78: check name is `manifest`

istae:
> > The label can be used strategically in general, because the GetCreateMutableBatch function matches the label, and it can reuse the same batch created before to save on cost and time (by not waiting for a new batch to be created which can take few minutes).
> 
> @istae Would this mean that we can resuse the same batch created with specific check? Or is it ok to use `test-label` for all of the checks? What is your suggestion, to leave the same label for all of the checks?

If the tests run one by one then it's more optimal to simply reuse the same batch.

pkg/check/manifest/manifest.go
```diff
@@ -86,57 +105,169 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
+// If the file parameter is nil, it downloads the index file in the collection.
+func (c *Check) download(client *bee.Client, address swarm.Address, file *bee.File, indexFile bee.File) error {
+	fName := ""
+	if file != nil {
+		fName = file.Name()
+	}
+	c.logger.Infof("downloading file: %s/%s", address, fName)
+
+	var hash []byte
+	for i := 0; i < 5; i++ {
```
istae, line 230: is some many retries necessary? is this just a precaution or you saw that it was failing quite often?

pkg/check/manifest/manifest.go
```diff
@@ -86,57 +105,169 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
...
+	}
+	return nil
+}
 
-		if !bytes.Equal(file.Hash(), hash) {
-			c.logger.Infof("Node %s. File %d not retrieved successfully. Uploaded size: %d Downloaded size: %d Node: %s File: %s/%s", lastNode, i, file.Size(), size, overlays[lastNode].String(), tarFile.Address().String(), file.Name())
-			return errManifest
+// download retrieves a file from the given address using the specified client.
+// If the file parameter is nil, it downloads the index file in the collection.
+func (c *Check) download(client *bee.Client, address swarm.Address, file *bee.File, indexFile bee.File) error {
```
istae, line 222: maybe rename downloadAndVerify

pkg/check/manifest/manifest.go
acha-bill: The retry count was always 5. I just refactored the code. I didn't want to mess with it.
https://github.com/ethersphere/beekeeper/blob/master/pkg/check/manifest/manifest.go#L115

istae (approved):

PR #454, opened 2025-01-31
feat: add PostageTTL and PostageLabel options to all Beekeeper checks

gacevicljubisa:


istae:
we should deprecate postage amount btw

darkobas2 (approved):

config/public-testnet.yaml
```diff
@@ -57,7 +57,7 @@ checks:
   pt-retrieval:
     options:
       chunks-per-node: 3
-      postage-amount: 140000000
+      postage-ttl: 24h
```
istae, line 60: since these batches are shared across the test because they have the same label, it's better to create large batches in terms of size and TLL. 
TTL should be two weeks, and depth 21 🤔

istae: also i don't see the postage-labels here?

cmd/beekeeper/cmd/cmd.go
```diff
@@ -27,21 +27,27 @@ import (
...
-	optionNameConfigGitRepo      = "config-git-repo"
-	optionNameConfigGitDir       = "config-git-dir"
 	optionNameConfigGitBranch    = "config-git-branch"
-	optionNameConfigGitUsername  = "config-git-username"
+	optionNameConfigGitDir       = "config-git-dir"
 	optionNameConfigGitPassword  = "config-git-password"
+	optionNameConfigGitRepo      = "config-git-repo"
+	optionNameConfigGitUsername  = "config-git-username"
+	optionNameEnableK8S          = "enable-k8s"
+	optionNameGethURL            = "geth-url"
```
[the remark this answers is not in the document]
gacevicljubisa: this would require changes in helm charts?

config/public-testnet.yaml
gacevicljubisa: > also i don't see the postage-labels here?

if not defined, default is `test-label`

version.go
```diff
@@ -1,7 +1,7 @@
 package beekeeper
 
 var (
-	version = "0.25.0" // manually set semantic version number
```
istae, line 4: this manual updating can be automated with something like this from Bee https://github.com/ethersphere/bee/blob/master/Makefile#L23

pkg/swap/block.go
```diff
@@ -18,12 +36,12 @@ func (g *GethClient) FetchBlockTime(ctx context.Context) (int64, error) {
 		return 0, fmt.Errorf("fetch latest block timestamp: %w", err)
 	}
 
-	timestampPrevious, err := g.fetchBlockTimestamp(ctx, latestBlockNumber-1)
+	timestampPrevious, err := g.fetchBlockTimestamp(ctx, latestBlockNumber-o.offset)
```
istae, line 39: what if latestBlockNumber is less than o.offset ? this could occur in custom environment with new deployments of geth.

istae: also by default o.offset is zero, the option to define it is not used.

istae (changes requested):

pkg/swap/block.go
gacevicljubisa: > what if latestBlockNumber is less than o.offset ? this could occur in custom environment with new deployments of geth.

Well, I was hoping that RPC will return error in this case. But maybe I can implement some retry logic that can half the offset until it finds the value? What do you think?

istae: i think using zero should be ok. 
you can do something like:
```
if latestBlockNumber < o.offset {
  g.fetchBlockTimestamp(ctx, 0)
}
```

gacevicljubisa: For some reason, when block 0 is used to query the Geth node, it returns the timestamp 1582023590, which corresponds to February 18, 2020. To avoid this issue, I limited the maximum offset to half of the latest block number.

istae (approved):
great work, thanks! 🥇

pkg/bee/client.go
```diff
@@ -38,14 +41,17 @@ type ClientOptions struct {
...
 
 // NewClient returns Bee client
 func NewClient(opts ClientOptions, log logging.Logger) (c *Client) {
 	c = &Client{
-		retry: retryCount,
-		opts:  opts,
-		log:   log,
+		retry:      retryCount,
+		log:        log,
+		swapClient: opts.SwapClient,
```
janos, line 52: In case that opts.SwapClient is nil, use swap.NotSet.

gacevicljubisa: thanks for the idea :)

janos (approved):
Nice! LGTM.

PR #455, opened 2025-02-06
chore: update pt-testnet-load check

gacevicljubisa:


config/public-testnet.yaml
```diff
@@ -131,21 +131,19 @@ checks:
   pt-testnet-load:
     options:
       content-size: 50000000
-      postage-amount: 1000000000 # 1024/15*60*60*24*7
+      postage-amount: 1428084000 # (28335/12)*60*60*24*7
       postage-depth: 24
-      duration: 12h
+      duration: 24h
```
istae, line 136: why not even longer

config/public-testnet.yaml
```diff
@@ -131,21 +131,19 @@ checks:
   pt-testnet-load:
     options:
       content-size: 50000000
-      postage-amount: 1000000000 # 1024/15*60*60*24*7
+      postage-amount: 1428084000 # (28335/12)*60*60*24*7
```
istae, line 134: hopefully we can drop this amount soon 😆

config/public-testnet.yaml
```diff
@@ -131,21 +131,19 @@ checks:
...
-      max-storage-radius: 3
+      max-storage-radius: 2
       upload-groups:
         - bee-1
         - bee-2
       download-groups:
         - bee-3
         - bee-4
-    timeout: 12h
-    type: load
```
istae: was this accidently removed?

config/public-testnet.yaml
gacevicljubisa: working on it :)

config/public-testnet.yaml
gacevicljubisa: feel free to suggest any values? When we deploy it as cron job, it will be restarted in any case?

istae: make it a month so we dont see so many pods created in k9s :)

istae (approved):

config/public-testnet.yaml
gacevicljubisa: To increase the postage-amount as well?

gacevicljubisa: To increase the postage-amount as well? To not cause some overspending?

gacevicljubisa: done

istae (approved):

PR #462, opened 2025-02-18
fix: more reliable pss and gsoc

acha-bill:
1. Add retry for expected pss messages
2. Refactor the WebSocket listener to not read on a closed connection

### DIFF
--- a/pkg/check/pss/pss.go
+++ b/pkg/check/pss/pss.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
-	"net/http"
 	"time"
 
 	"github.com/ethersphere/beekeeper/pkg/bee"
@@ -12,8 +11,7 @@ import (
 	"github.com/ethersphere/beekeeper/pkg/logging"
 	"github.com/ethersphere/beekeeper/pkg/orchestration"
 	"github.com/ethersphere/beekeeper/pkg/random"
-	websocket2 "github.com/ethersphere/beekeeper/pkg/websocket"
-	"github.com/gorilla/websocket"
+	"github.com/ethersphere/beekeeper/pkg/wslistener"
 )
 
 // Options represents check options
@@ -123,7 +121,7 @@ func (c *Check) testPss(nodeAName, nodeBName string, clients map[string]*bee.Cli
 	}
 	c.logger.Infof("node %s: batched id %s", nodeAName, batchID)
 
-	ch, closer, err := listenWebsocket(ctx, nodeB.Host(), testTopic, c.logger)
+	ch, closer, err := wslistener.ListenWebSocket(ctx, nodeB, fmt.Sprintf("/pss/subscribe/%s", testTopic), c.logger)
 	if err != nil {
 		return err
 	}
@@ -137,6 +135,7 @@ func (c *Check) testPss(nodeAName, nodeBName string, clients map[string]*bee.Cli
 	if err != nil {
 		return err
 	}
+	c.logger.Infof("pss: test data sent successfully to node %s. Waiting for response from node %s", nodeAName, nodeBName)
 
 L:
 	for {
@@ -160,18 +159,4 @@ L:
 	}
 
 	return err
-}
-
-func listenWebsocket(ctx context.Context, host string, topic string, logger logging.Logger) (<-chan string, func(), error) {
-	dialer := &websocket.Dialer{
-		Proxy:            http.ProxyFromEnvironment,
-		HandshakeTimeout: 45 * time.Second,
-	}
-
-	ws, _, err := dialer.DialContext(ctx, fmt.Sprintf("ws://%s/pss/subscribe/%s", host, topic), http.Header{})
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return websocket2.ListenWebSocket(ctx, ws, logger)
 }
--- a/pkg/check/pss/pss.go
+++ b/pkg/check/pss/pss.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ethersphere/beekeeper/pkg/logging"
 	"github.com/ethersphere/beekeeper/pkg/orchestration"
 	"github.com/ethersphere/beekeeper/pkg/random"
+	websocket2 "github.com/ethersphere/beekeeper/pkg/websocket"
 	"github.com/gorilla/websocket"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/websocket/websocket.go
```diff
@@ -0,0 +1,64 @@
+package websocket
```
gacevicljubisa, line 1: Nice idea to extract this to package. Maybe some different name, like `wslistener`, or `wsstream` if we will add more streaming related functionalities.

pkg/websocket/websocket.go
```diff
@@ -0,0 +1,64 @@
...
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	closer := func() {
+		close(done)
+		ws.Close()
+	}
```
gacevicljubisa, line 61: It must be ensured that `closer` is called just once.

pkg/websocket/websocket.go
```diff
@@ -0,0 +1,64 @@
...
+			_, data, err := ws.ReadMessage()
+			if err != nil {
+				errCh <- err
+				return
+			}
+			select {
+			case readCh <- data:
+			case <-done:
+				return
+			}
```
gacevicljubisa, line 31: Check for `<-ctx.Done()` is required in select

pkg/websocket/websocket.go
```diff
@@ -0,0 +1,64 @@
...
+	readCh := make(chan []byte)
+	errCh := make(chan error)
+	done := make(chan struct{})
+
+	go func() {
+		defer close(readCh)
+		defer close(errCh)
+
+		for {
+			_, data, err := ws.ReadMessage()
```
gacevicljubisa, line 22: It is possible to use ws.SetReadDeadline to prevent blocking forever.

pkg/check/pss/pss.go
```diff
@@ -100,71 +100,66 @@ func pickAtRandom(r *rand.Rand, names []string, skip string) string {
...
 
 	batchID, err := nodeA.GetOrCreateMutableBatch(ctx, o.PostageTTL, o.PostageDepth, o.PostageLabel)
 	if err != nil {
-		cancel()
 		return fmt.Errorf("node %s: batched id %w", nodeAName, err)
 	}
 	c.logger.Infof("node %s: batched id %s", nodeAName, batchID)
 
-	ch, close, err := listenWebsocket(ctx, nodeB.Host(), testTopic, c.logger)
+	ch, closer, err := listenWebsocket(ctx, nodeB.Host(), testTopic, c.logger)
```
gacevicljubisa, line 126: It would be nice that we log information which node is listening.

pkg/wslistener/wslistener.go
```diff
@@ -0,0 +1,97 @@
...
+	errCh := make(chan error)
+	done := make(chan struct{})
+
+	go func() {
+		defer close(readCh)
+		defer close(errCh)
+
+		for {
+			// ping should be received every 60s (defined on bee)
+			err := ws.SetReadDeadline(time.Now().Add(90 * time.Second))
```
martinconic, line 38: 60s in the comment (defined on bee) and 90s in the code?

pkg/wslistener/wslistener.go
```diff
@@ -0,0 +1,97 @@
...
+			if err != nil {
+				errCh <- err
+				return
```
martinconic, line 41: Shouldn't you also return the error? Same bellow

acha-bill: The err is passed to the error channel

pkg/wslistener/wslistener.go
acha-bill: Ping is expected every 60s. If we wait for up to 90s, then timeout.

pkg/wslistener/wslistener.go
```diff
@@ -0,0 +1,97 @@
...
+			select {
+			case data := <-readCh:
+				logger.WithField("node", client.Name()).Infof("websocket received message: %s", string(data))
+				select {
+				case ch <- string(data):
+				case <-done:
+					return
+				}
+			case err := <-errCh:
+				logger.Infof("websocket error: %v", err)
```
gacevicljubisa, line 71: use `logger.Errorf`

pkg/wslistener/wslistener.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}()
+
+	var once sync.Once
+	closer := func() {
+		once.Do(func() {
+			deadline := time.Now().Add(5 * time.Second)
+			msg := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+			if err := ws.WriteControl(websocket.CloseMessage, msg, deadline); err != nil {
+				logger.Infof("failed to send close message: %v", err)
```
gacevicljubisa, line 88: also, use `logger.Errorf`

pkg/wslistener/wslistener.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			close(done)
+			ws.Close()
```
gacevicljubisa, line 92: Maybe it is better first to close the connection and than to close channel, also, `ws.Close()` returns an error, so why not log it if it happens.

gacevicljubisa:
For example, if we do not defer `closer()` in pss.go, and call it right after `wslistener.ListenWebSocket`, check will complete with success. But we didn't receive any response, because of closed connection. The check must ensure that we receive correct data.

pkg/wslistener/wslistener.go
gacevicljubisa: By using the erorr level, we can setup alerts on Grafana when they occur.

martinconic (approved):

janos (commented):
I agree with the remaining comments, otherwise, all is good.

pkg/wslistener/wslistener.go
acha-bill: closing the socket first can result in a `read on closed connection`.
So we close the channel first to ensure that the goroutine returns. Then close the underlying connection

pkg/check/pss/pss.go
```diff
@@ -100,96 +98,61 @@ func pickAtRandom(r *rand.Rand, names []string, skip string) string {
...
-			c.logger.Info("pss: websocket connection received correct message")
-			c.metrics.SendAndReceiveGauge.WithLabelValues(nodeAName, nodeBName).Set(time.Since(tStart).Seconds())
-		} else {
-			err = errDataMismatch
+	for {
+		select {
+		case <-time.After(1 * time.Minute):
+			return fmt.Errorf("correct message not received after %s", 1*time.Minute)
+		default:
+			msg, ok := <-ch
```
gacevicljubisa, line 145: Select will never block, and it will never reach `time.After`, because every loop it will fallthrough to default case. Maybe something like this will work better:
```
		select {
		case <-time.After(1 * time.Minute):
			return fmt.Errorf("correct message not received after %s", 1*time.Minute)
		case msg, ok := <-ch:
			if !ok {
				return fmt.Errorf("ws closed before receiving correct message")
			}

			stringData := string(testData)

			if msg != stringData {
				c.logger.Warningf("pss: received incorrect message. trying again. want %s, got %s", stringData, msg)
			}
			c.logger.Info("pss: websocket connection received correct message")
			c.metrics.SendAndReceiveGauge.WithLabelValues(nodeAName, nodeBName).Set(time.Since(tStart).Seconds())
			return nil
		}
		```

gacevicljubisa (approved):

PR #463, opened 2025-02-19
fix(check-runner): improve error aggregation for multiple check failures

gacevicljubisa:
This commit improves the error handling in the check-runner by refining how errors are managed and reported:

- When a single check is executed and an error occurs, the error is returned immediately.
- When multiple checks are executed, every check is run regardless of errors. Each error is logged, and after all checks have run, a summary of all errors is returned.

### DIFF
--- a/pkg/check/runner.go
+++ b/pkg/check/runner.go
@@ -87,7 +87,7 @@ func (c *CheckRunner) Run(ctx context.Context, checks []string) error {
 			typeName: checkConfig.Type,
 			action:   chk,
 			options:  o,
-			timeout:  *checkConfig.Timeout,
+			timeout:  checkConfig.Timeout,
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


janos (approved):

pkg/check/runner.go
```diff
@@ -0,0 +1,168 @@
...
+	if len(errors) == 1 {
+		return errors[0].err
+	} else if len(errors) > 1 {
+		var errStrings []string
+		for _, e := range errors {
+			errStrings = append(errStrings, fmt.Sprintf("[%s]: {%v}", e.check, e.err))
+			c.logger.Errorf("%s: %v", e.check, e.err)
+		}
+		return fmt.Errorf("multiple checks failed: %s", strings.Join(errStrings, "; "))
+	}
```
acha-bill, line 125: cosmetic: you don't need else after a return

acha-bill (approved):

PR #465, opened 2025-02-27
fix: gsoc via pushsync

acha-bill:
A SOC can arrive at a node via push sync or pull sync.
With pull sync, you might not get all chunk versions because the newer version would have overridden the older one on the upstream node by the time you sync. 

This issue doesn't exist when the listener node receives the chunk via pushsync as it will receive the chunk immediately after it is pushed. Effectively receiving all versions.

This PR ensures that the listener gets the chunk via push sync and such, we can verify consistently that it receives all versions.

### DIFF
--- a/pkg/orchestration/k8s/cluster.go
+++ b/pkg/orchestration/k8s/cluster.go
@@ -443,3 +443,43 @@ func (c *Cluster) FlattenTopologies(ctx context.Context) (topologies map[string]
 
 	return
 }
+
+// ClosetFullNodeClient returns the closest full node client to the supplied node.
+func (c *Cluster) ClosetFullNodeClient(ctx context.Context, s *bee.Client, r *rand.Rand) (*bee.Client, error) {
+	addrToNode := make(map[string]orchestration.Node)
+	for _, n := range c.Nodes() {
+		res, err := n.Client().Addresses(ctx)
+		if err != nil {
+			return nil, err
+		}
+		addrToNode[res.Overlay.String()] = n
+	}
+
+	t, err := s.Topology(ctx)
+	if err != nil {
+		return nil, err
+	}
+	const maxBin = 32
+	for b := range maxBin {
+		bin := t.Bins[fmt.Sprintf("bin_%d", b)]
+		var fullNodes []orchestration.Node
+		for _, peer := range bin.ConnectedPeers {
+			node, ok := addrToNode[peer.Address.String()]
+			if !ok {
+				return nil, fmt.Errorf("peer overlay %s not found in address map", peer.Address.String())
+			}
+			cfg := node.Config()
+			if cfg.FullNode && !cfg.BootnodeMode {
+				fullNodes = append(fullNodes, node)
+			}
+		}
+		if len(fullNodes) == 0 {
+			continue
+		}
+		r.Shuffle(len(fullNodes), func(i, j int) {
+			fullNodes[i], fullNodes[j] = fullNodes[j], fullNodes[i]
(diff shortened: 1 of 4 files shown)


pkg/orchestration/k8s/cluster.go
```diff
@@ -443,3 +443,43 @@ func (c *Cluster) FlattenTopologies(ctx context.Context) (topologies map[string]
 
 	return
 }
+
+// ClosetFullNodeClient returns the closest full node client to the supplied node.
+func (c *Cluster) ClosetFullNodeClient(ctx context.Context, s *bee.Client, r *rand.Rand) (*bee.Client, error) {
```
gacevicljubisa, line 448: Should be `Closest`, it is a typo

gacevicljubisa: Is it possible to have `ClosestFullNode` method on a bee.Client?

janos: I agree, it would be logical that bee client provides closest peer address and here to return the client.

pkg/orchestration/k8s/cluster.go
```diff
@@ -443,3 +443,43 @@ func (c *Cluster) FlattenTopologies(ctx context.Context) (topologies map[string]
...
+			}
+			cfg := node.Config()
+			if cfg.FullNode && !cfg.BootnodeMode {
+				fullNodes = append(fullNodes, node)
+			}
+		}
+		if len(fullNodes) == 0 {
+			continue
+		}
+		r.Shuffle(len(fullNodes), func(i, j int) {
```
janos, line 479: Instead of picking a random peer, would `swarm.Proximity` function provide the actual closest peer out of all in the closes bin?

pkg/orchestration/k8s/cluster.go
acha-bill: The client cannot return another node/client. It also cannot tell if a peer is a full node or not (without an extra API call)
It can return the address of the closest peer but this peer might not be a full node. So this is not reliable.
IMO, `cloestFullNode` should be in the orchestration package.

janos: If /topology would include the full node status for every connected peer something like bins.bin_xx.connectedPeers[].fullNode, that would be quite handy.

The bee client should return only the address of the closes peer, and the client for that peer can be created wherever is needed.

In any case it is not a crucial thing, it would just make a better separation of concerns.

acha-bill: To achieve this without making any API changes, `client.ClosestPeer` can take in a `bin` and will return the closest peer in that bin. Then the caller can iteratively find the closest full node starting at bin 0.
Would this make more sense?

janos: That is a very nice idea, like an iterator over peers by proximity. Sound good to me. I would be interested to hear @gacevicljubisa as well, as he has found this improvement.

gacevicljubisa: My suggestion was just because separation of concerns, implmentation is yours and it is fine by me. :)

I guess that we could extract bee interface to specific package and then be able to achive that Client can return another one. But this would require refactoring. This we can do in the future. 

Also, just to mention, we have one similar function [ClosestNodeFromMap](https://github.com/ethersphere/beekeeper/blob/a2803bfd744959195139e1f5fea5726bb663322b/pkg/bee/chunk.go#L112).

gacevicljubisa (approved):

PR #467, opened 2025-03-03
fix: gsoc listner node selection

acha-bill:
Increase the depth for mining the soc addresses so that the listener node is picked more accurately when pushing out the chunk on the uploader.

Also increases the timeout to receive all messages.

### DIFF
--- a/pkg/check/gsoc/gsoc.go
+++ b/pkg/check/gsoc/gsoc.go
@@ -121,9 +121,9 @@ func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client
 	if err != nil {
 		return err
 	}
-	depth := 6
-	logger.Infof("gsoc: mining resource id for overlay=%s, depth=%d", addresses.Overlay, depth)
-	resourceId, socAddress, err := mineResourceId(ctx, addresses.Overlay, privKey, depth)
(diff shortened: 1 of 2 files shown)


pkg/check/gsoc/gsoc.go
```diff
@@ -121,7 +121,9 @@ func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client
 	if err != nil {
 		return err
 	}
-	resourceId, socAddress, err := mineResourceId(ctx, addresses.Overlay, privKey, 1)
+	depth := 6
```
gacevicljubisa, line 124: Why depth of 6?

istae: maybe the depth should be configurable, so we can 1 for the CI, and and a different value for the testnet?

istae: for depth, maybe commitment depth can be used from the /status endpoint?

gacevicljubisa: It can be retrieved from [here](https://github.com/ethersphere/beekeeper/blob/81c573a38b723b1b1b16d176948e3a8c47d93817/pkg/bee/client.go#L1022)

martinconic (approved):

PR #478, opened 2025-03-21
fix: fix the wrong error return value

wellweek:
Since we have already checked err before and returned when err != nil, err must be nil here. In fact, it should return notGErr.


pkg/check/act/act.go
```diff
@@ -261,5 +261,5 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts int
 		return fmt.Errorf("node %s: File downloaded with wrong public key successfully - this is an error", nodeName2)
 	}
 	c.logger.Info("ACT Access denied for not enabled grantee after patch")
-	return
+	return notGErr
```
gacevicljubisa, line 264: In this case `notGErr` is not `nil`, and if we return error, check will fail. Intention of the code, DownloadActFile on line 259 should return error for `notGErr`. Then it would be best that we check the contents of the error, and return `nil`.

gacevicljubisa (changes requested):

PR #505, opened 2025-08-08
feat: smoke file sizes

nugaon:


config/local.yaml
```diff
@@ -282,7 +282,7 @@ checks:
     type: settlements
   ci-smoke:
     options:
-      content-size: 50000
+      file-sizes: [3072, 30720, 307200, 31457280, 104857600, 314572800] # [3KB, 30KB, 300KB, 30MB, 100MB, 300MB]
```
gacevicljubisa, line 285: For the local.yaml, keep some smaller sizes, maybe max to 3 mb. 
Local deployments do not have persistant-storage supported.

config/local.yaml
```diff
@@ -295,7 +295,7 @@ checks:
     type: smoke
   ci-load:
     options:
-      content-size: 50000
+      file-sizes: [31457280, 104857600, 314572800] # [30MB, 100MB, 300MB]
```
gacevicljubisa, line 298: For the local.yaml, keep some smaller sizes, maybe max to 3 mb.
Local deployments do not have persistant-storage supported.

gacevicljubisa:
Code changes are looking good to me. We can execute this locally agianst bee-light-testnet to verify. 
Also, changes are required in yaml files here as well https://github.com/ethersphere/beekeeper-config/tree/master/config

config/config.yaml
```diff
@@ -335,7 +335,7 @@ checks:
     type: settlements
   smoke:
     options:
-      content-size: 5000
+      file-sizes: [31457280, 104857600, 314572800] # [30MB, 100MB, 300MB]
```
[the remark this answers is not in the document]
gacevicljubisa: In some deployments, the latest tag is used, while others specify a particular version. Since configurations are shared, we should retain the `content-size` option in these configs as a precaution until we're certain it's safe to remove. It may be fine to remove the content-size option from the configs in this repository. Additional configuration locations might exist in @darkobas2 repositories. To ensure backward compatibility, I suggest supporting both options, as the fix should be minimal. Worse case scenario is that smoke job will not work, and we need to ensure that configurations and version are updated.

PR #508, opened 2025-08-13
chore: bump golangci-lint to v2 and golang to v1.25.1

gacevicljubisa:


akrem-chabchoub (approved):
LGTM. 
Upgrade looks great. Would it make sense to enable `gosec` and `gocognit` for improved security and complexity checks?

PR #516, opened 2025-08-22
refactor: enhance restart command to utilize NodeProvider

akrem-chabchoub:
# Refactor: Use NodeProvider interface for restart command

## Problem
The restart command was tightly coupled to the orchestration package, making it difficult to reuse and maintain. It had separate methods for cluster and namespace-based restarts with duplicated logic.

## Solution
- Refactored to use NodeProvider interface for unified node discovery
- Simplified restart client with single `Restart()` method
- Added node group filtering capabilities
- Added direct pod restart support for namespace-based operations
- Improved error handling and logging

## Changes
- **Restart Client Refactoring**: Replaced orchestration dependency with NodeProvider interface
- **Node Group Filtering**: Added support for filtering nodes by groups in beekeeper deployments
- **Command Simplification**: Unified cluster and namespace restart logic
- **Pod-Level Restart**: Added direct pod restart using label selectors for namespace operations
- **Deployment Type Support**: Added `--deployment-type` flag for beekeeper vs helm deployments

## Usage Examples

```bash
# Restart with new image
beekeeper restart --cluster-name=my-cluster --image=bee:v2.0.0

# Namespace-based restart with pod label selector
beekeeper restart --namespace=my-namespace --label-selector="type:type1"
```

Closes https://github.com/ethersphere/beekeeper/issues/526


cmd/beekeeper/cmd/restart.go
```diff
@@ -12,60 +12,65 @@ import (
...
-				}
+				if clusterName != "" {
+					clusterConfig, ok := c.config.Clusters[clusterName]
+					if !ok {
+						return fmt.Errorf("cluster config %s not defined", clusterName)
+					}
 
-				return nil
-			}
+					cluster, err := c.setupCluster(ctx, clusterName, false)
```
gacevicljubisa, line 49: `c.setupCluster(ctx, clusterName, false)` is already called in `c.createNodeClient(ctx, true)`, no need to call it twice.

gacevicljubisa:
If image is not specified, pod should only be restarted. If specified, then update the Image of the statefull set and then pod should be deleted if strategy is `UpdateStrategyOnDelete`

pkg/node/client.go
```diff
@@ -193,6 +207,51 @@ func (sc *Client) getIngressNodes(ctx context.Context) ([]Node, error) {
 	return nodes, nil
 }
 
+// filterClientsByNodeGroups filters bee clients by node groups
+// Node names in beekeeper deployments follow the pattern: {nodeGroup}-{index}
+// This function extracts the node group from the node name and filters accordingly
+func (sc *Client) filterClientsByNodeGroups(clients map[string]*bee.Client) map[string]*bee.Client {
```
gacevicljubisa, line 213: You do not need this logic. Much simplier and more precise solution would be to add `NodeGroupName` in `bee.ClientOptions`, and to expose new method NodeGroup() in `bee` package. Then you can compare on name. This should be done only in case of beekeeper deployment.

gacevicljubisa: This is psudo code how this filtering could work:
```go
	filteredClients := make(map[string]*bee.Client)
	if len(sc.nodeGroups) > 0 && sc.deploymentType == DeploymentTypeBeekeeper {
		for _, ng := range sc.nodeGroups {
			for name, client := range sc.beeClients {
				if client.NodeGroup() == ng {
					filteredClients[name] = client
				}
			}
		}
	} else {
		filteredClients = sc.beeClients
	}
```

Issue #523, opened 2025-09-09
Refactor: Separate smoke and load checks into distinct packages


gacevicljubisa:
## Problem
Currently, both the basic smoke test and load test are implemented in the same `pkg/check/smoke` package, which creates confusion and makes the codebase harder to maintain. These two checks serve different purposes:

- **Smoke Check**: Basic functionality testing with single uploader/downloader, multiple file sizes, and simple validation
- **Load Check**: Performance/stress testing with concurrent uploaders/downloaders, committed depth management, and load-specific features

## Proposed Solution
Separate these checks into two distinct packages, and extract shared metrics where appropriate. Check if they can be reused in other `checks`.

gacevicljubisa:
@nugaon feel free to add contribution to this issue

PR #525, opened 2025-09-26
refactor: separate smoke and load checks into distinct packages

akrem-chabchoub:
# Refactor: Separate smoke and load checks into distinct packages

## Problem
Both smoke and load tests were implemented in the same `pkg/check/smoke` package, creating confusion and making maintenance difficult.

## Solution
- ✅ Separated into `pkg/check/smoke/` and `pkg/check/load/` packages
- ✅ Improved logging messages

## Changes
- **Load and Smoke Check Refactoring**: Separate them in 2 packages
- **Configuration**: Both checks properly registered in config system
- **Metrics**: Separate metrics collection for each check type

Closes [#523](https://github.com/ethersphere/beekeeper/issues/523)

### DIFF
--- a/pkg/check/load/load.go
+++ b/pkg/check/load/load.go
@@ -191,15 +191,15 @@ func (c *Check) run(ctx context.Context, cluster orchestration.Cluster, o Option
 
 					c.metrics.UploadAttempts.WithLabelValues(sizeLabel).Inc()
 					var duration time.Duration
-					c.logger.Infof("uploading to: %s", uploader)
+					c.logger.Infof("uploading to: %s", uploader.Name())
 
 					batchID, err := uploader.GetOrCreateMutableBatch(ctx, o.PostageTTL, o.PostageDepth, o.PostageLabel)
 					if err != nil {
-						c.logger.Errorf("create new batch: %v", err)
+						c.logger.Errorf("create new batch failed: %v", err)
 						return
 					}
 
-					c.logger.WithField("batch_id", batchID).Info("using batch")
+					c.logger.WithField("batch_id", batchID).Infof("node %s: using batch", uploader.Name())
 
 					address, duration, err = test.Upload(ctx, uploader, txData, batchID)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/check/load/load.go
```diff
@@ -12,35 +12,68 @@ import (
...
 	metrics metrics
 	logger  logging.Logger
 }
 
-// NewCheck returns new check
-func NewLoadCheck(log logging.Logger) beekeeper.Action {
-	return &LoadCheck{
+func NewCheck(log logging.Logger) beekeeper.Action {
+	return &Check{
 		metrics: newMetrics("check_load"),
```
gacevicljubisa, line 71: If the metrics are completly the same as in smoke, we can expose those in smoke, and initialize them here `smoke.NewMetrics("check_load")`.

pkg/check/load/load.go
```diff
@@ -270,6 +299,39 @@ func (c *LoadCheck) checkCommittedDepth(ctx context.Context, client *bee.Client,
 	}
 }
 
+type test struct {
```
gacevicljubisa, line 302: This is also already existing in smoke, and we can expose it there and use it in load.

gacevicljubisa:
I would suggest using the `ShuffledFullNodeClients` where possibile in `load` and `smoke`, because both checks have their own logic how to pick random client, but this is already imeplemented [here](https://github.com/ethersphere/beekeeper/blob/c1ca7cfa0225f58c8223bb13571809d689f10f38/pkg/orchestration/k8s/cluster.go#L251) and you can just use any of them for first for download and second for upload. You can also check how it is used in some other checks. This will allow you to refactor and improve both checks.

pkg/check/load/load.go
gacevicljubisa: We do not have repeated code, metrics were the same. Now we have different labels, so it is ok to split them com completely.

gacevicljubisa (approved):

Issue #527, opened 2025-09-30
Performance Benchmark: Impact of Chequebook Funding on Upload & Download Performance


akrem-chabchoub:
# Performance Benchmark: Impact of Chequebook Funding on Upload & Download Performance

## Description

We aim to assess the impact of chequebook funding on upload and download performance in the Bee network.

✅ **This benchmark can be implemented using the existing smoke check in Beekeeper, which already provides all the required functionality.**

### Example configuration for local testing:
`beekeeper check --cluster-name local-dns --checks performance-benchmark --config config/local.yaml`
```yaml
checks:
  performance-benchmark:
    options:
      file-sizes: [1024, 10240, 102400, 1048576, 52428800, 209715200] # [1KB, 10KB, 100KB, 1MB, 50MB, 200MB]
      postage-ttl: 24h
      postage-depth: 21
      postage-label: performance-test
      nodes-sync-wait: 30s
      upload-timeout: 7m
      download-timeout: 7m
      iteration-wait: 2m
      duration: 14m  
    timeout: 15m
    type: smoke
```

## Requirements

Measure upload and download times under two conditions:

- **With chequebook funded**
- **Without chequebook funded**

You can configure this in the bee config.yml:
```
chequebook-enable: false
```

### File sizes for testing:
- 1 KB
- 10 KB
- 100 KB
- 1 MB
- 50 MB
- 200 MB

> **Note:** Ensure that chunks are retrieved from the network and not served from local cache.

### Testing setup:
- Use the mainnet
- Ideally, use a cluster with two separate nodes (one for upload, one for download) with chequebook disabled.
- Alternatively, static endpoints can be used if already existing reachable nodes have chequebook disabled.

## Implementation

All of this functionality is already implemented in the smoke check, so no additional custom implementation is required. You can use smoke check configuration options to run the benchmark and collect measurements.

## Expected Outcome

Determine whether chequebook funding introduces additional latency in file transfers, guiding future research directions.

gacevicljubisa:
Beekeeper smoke metrics are already updated and they should have label with file sizes added to them.

crtahlin:
Who is the assignee for the next step in this issue, what is the next step? @gacevicljubisa , @darkobas2 ?

crtahlin:
@akrem-chabchoub is above as expected, can the issue be closed?

PR #529, opened 2025-10-04
check-smoke: add node name label to metrics

darkobas2:
since https://github.com/ethersphere/beekeeper/issues/527 requires thow different configurations to tests we should add node_name label to metrics so we can distingish results in metrics. we get the info from beekeeper config


cmd/beekeeper/cmd/metrics.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 // newMetricsPusher returns a new metrics pusher and a cleanup function.
-func newMetricsPusher(pusherAddress, job string, logger logging.Logger) (*push.Pusher, func()) {
-	metricsPusher := push.New(pusherAddress, job)
+func newMetricsPusher(pusherAddress, job, podName string, logger logging.Logger) (*push.Pusher, func()) {
+	metricsPusher := push.New(pusherAddress, job).Grouping("instance", podName)
```
gacevicljubisa, line 14: In case that beekeeper is not deployed in cluster, I would suggest fallback to hostname:

```go
func newMetricsPusher(pusherAddress, job string, logger logging.Logger) (*push.Pusher, func()) {
    instanceName := getInstanceName()
    metricsPusher := push.New(pusherAddress, job).Grouping("instance", instanceName)
    // ... rest of function
}

func getInstanceName() string {
    if podName := os.Getenv("POD_NAME"); podName != "" {
        return podName
    }
    if hostname, err := os.Hostname(); err == nil {
        return hostname
    }
    return "unknown"
}
```
You can resolve instance name internally in constructor.

gacevicljubisa (approved):

cmd/beekeeper/cmd/metrics.go
darkobas2: Ah yeah, should be Hostname all together i dont think pod_name even exists

akrem-chabchoub (approved):

cmd/beekeeper/cmd/metrics.go
gacevicljubisa: We can set it as env explicitly? 
Real pod name would not be maybe appropriate, because every time when pod is recreated it gets uuid on top of the cron job name. But we can set it explicitly during deployment, to have the same name as cron job?

darkobas2: The pod_name change i forgot to remove, that was first idea but then i figured we can just use node_group (labelNodeName)  from config right?

So i will revert that

PR #540, opened 2025-10-17
feat(nuke): add option for nuke of stateful sets by name

gacevicljubisa:


pkg/nuker/nuker.go
```diff
@@ -105,6 +105,42 @@ func (c *Client) Run(ctx context.Context, restartArgs []string) (err error) {
...
+	c.log.Info("starting Bee cluster nuke by StatefulSet names")
+
+	if len(statefulSetNames) == 0 {
+		return errors.New("stateful set names cannot be empty")
+	}
+
+	statefulSetsMap := make(map[string]*v1.StatefulSet)
+
+	for _, name := range statefulSetNames {
+		statefulSet, err := c.k8sClient.StatefulSet.Get(ctx, name, namespace)
```
akrem-chabchoub: When passing a non-existent StatefulSet, the command triggers a panic error:
`./dist/beekeeper nuke --namespace=local --stateful-sets="non-existant-bootnode" --restart-args="bee,start,--config=.bee.yaml"`

In this case, the `statefulSet` is `nil`, which causes the panic.

PR #544, opened 2025-10-30
chore: modify bee-playground-cluster

gacevicljubisa:


config/testnet-bee-playground.yaml
```diff
@@ -24,36 +33,19 @@ node-groups:
   ng-bee-playground:
     _inherit: default
     persistence-enabled: true
-    image: ethersphere/bee:latest
+    image: ljubisagacevic/bee:v37
```
akrem-chabchoub: This image is just for testing afaik ? so maybe we just add a todo to revert it ?

PR #557, opened 2026-01-02
fix: gsoc and pss checks

martinconic:
The pt-pss check was failing intermittently with "500 Internal Server Error" or timeouts, even when messages were successfully delivered. This PR adds retry logic and decouples the send confirmation from the message receipt verification.

The pt-gsoc check was timing out under high load when sending many chunks in parallel. This PR introduces a configurable Chunks option, reduces the default parallel chunk count, and implements a smart retry mechanism that only re-broadcasts missing chunks.

**PSS Check (pkg/check/pss)**

**Reliability**: Added retry logic (5 attempts) to SendPSSMessage.
**False Positive Fix**: Modified the check to succeed if the PSS message is received by the listener, even if the SendPSSMessage call returns an error (e.g., timeout).
**Configuration**: Updated default PostageDepth to 22 to align with testnet configuration.
GSOC Check (pkg/check/gsoc)

**Reliability**: Implemented a retry mechanism that detects missing SOC updates after the initial run and re-broadcasts only the missing chunks.
**Context Safety**: Added ctx.Done() checks in the retry loop to ensure the check respects context cancellation immediately.
**Configuration**:
Added a Chunks option to control the number of parallel updates (default: 3).
Updated default PostageDepth to 22.
Added strict validation: Returns an error if Chunks is ≤ 0 (previously defaulted silently to 3).

### DIFF
--- a/pkg/check/gsoc/gsoc.go
+++ b/pkg/check/gsoc/gsoc.go
@@ -29,14 +29,16 @@ type Options struct {
 	PostageTTL   time.Duration
 	PostageDepth uint64
 	PostageLabel string
+	Chunks       int
 }
 
 // NewDefaultOptions returns new default options
 func NewDefaultOptions() Options {
 	return Options{
 		PostageTTL:   24 * time.Hour,
-		PostageDepth: 17,
+		PostageDepth: 22,
 		PostageLabel: "test-label",
+		Chunks:       3,
 	}
 }
 
@@ -95,14 +97,14 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts any
 	}
 
 	c.logger.Infof("send messages with different postage batches sequentially...")
-	err = run(ctx, uploadClient, listenClient, batches, c.logger, false)
+	err = run(ctx, uploadClient, listenClient, batches, c.logger, false, o.Chunks)
 	if err != nil {
 		return fmt.Errorf("sequential: %w", err)
 	}
 	c.logger.Infof("done")
 
 	c.logger.Infof("send messages with different postage batches parallel...")
-	err = run(ctx, uploadClient, listenClient, batches, c.logger, true)
+	err = run(ctx, uploadClient, listenClient, batches, c.logger, true, o.Chunks)
 	if err != nil {
 		return fmt.Errorf("parallel: %w", err)
 	}
@@ -111,8 +113,10 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts any
 	return nil
 }
 
-func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client, batches []string, logger logging.Logger, parallel bool) error {
-	const numChunks = 10
+func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client, batches []string, logger logging.Logger, parallel bool, numChunks int) error {
+	if numChunks <= 0 {
+		numChunks = 3
+	}
 	privKey, err := crypto.GenerateSecp256k1Key()
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/check/gsoc/gsoc.go
```diff
@@ -111,8 +113,10 @@ func (c *Check) Run(ctx context.Context, cluster orchestration.Cluster, opts any
 	return nil
 }
 
-func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client, batches []string, logger logging.Logger, parallel bool) error {
-	const numChunks = 10
+func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client, batches []string, logger logging.Logger, parallel bool, numChunks int) error {
+	if numChunks <= 0 {
+		numChunks = 3
```
akrem-chabchoub: Should show a warning to user that we are changing number of chunks to 3 ?
Why 3 exactly ?

akrem-chabchoub (approved):

pkg/check/pss/pss.go
```diff
@@ -131,8 +131,38 @@ func (c *Check) testPss(nodeAName, nodeBName string, clients map[string]*bee.Cli
...
+		}
+
+		// check if message is received
+		select {
+		case msg := <-ch:
+			if msg == string(testData) {
+				c.logger.Info("pss: message received despite send failure")
+				return nil
+			}
+		default:
```
aloknerurkar: There seems to be a race here. We do a non-blocking check to see if we get a message and then retry after delay. Message can arrive while we are sleeping.

Better would be
```go
select {
case msg, ok := <-ch:
    if !ok {
        return errors.New("pss websocket closed")
    }
    if msg == string(testData) {
        return nil
    }

case <-time.After(time.Second):
    // Retry.

case <-ctx.Done():
    return ctx.Err()
}
```

pkg/check/pss/pss.go
```diff
@@ -131,8 +131,38 @@ func (c *Check) testPss(nodeAName, nodeBName string, clients map[string]*bee.Cli
...
+			// continue
+		}
+
+		c.logger.Infof("pss: send failed, retrying in 1s: %v", err)
+		time.Sleep(1 * time.Second)
```
aloknerurkar: One more reason for the above suggestion is that there could be a context cancellation that happens while we are sleeping and we will ignore that.

pkg/check/gsoc/gsoc.go
```diff
@@ -172,20 +174,21 @@ func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client
...
-		if !received[want] {
-			return fmt.Errorf("message '%s' not received", want)
+		return nil
+	case <-time.After(3 * time.Minute):
+		receivedMtx.Lock()
+		defer receivedMtx.Unlock()
+		for i := range numChunks {
+			want := fmt.Sprintf("data %d", i)
+			if !received[want] {
+				return fmt.Errorf("message '%s' not received", want)
```
aloknerurkar: This will return at the first message that was not received. There could be multiple, so it can be misleading error. Instead we should use errors.Join and report all the missing messages.

pkg/check/gsoc/gsoc.go
```diff
@@ -179,12 +180,16 @@ func run(ctx context.Context, uploadClient *bee.Client, listenClient *bee.Client
...
 		for i := range numChunks {
 			want := fmt.Sprintf("data %d", i)
 			if !received[want] {
-				return fmt.Errorf("message '%s' not received", want)
+				errs = append(errs, fmt.Errorf("message '%s' not received", want))
 			}
 		}
+		if len(errs) > 0 {
+			return errors.Join(errs...)
+		}
```
aloknerurkar: On second thought, I dont think errors.Join is actually the right thing here. Maybe better to just do string concatenation.
```go
var missing []string
for i := range numChunks {
    want := fmt.Sprintf("data %d", i)
    if !received[want] {
        missing = append(missing, want)
    }
}
if len(missing) > 0 {
    return fmt.Errorf("messages not received: %s", strings.Join(missing, ", "))
}
```

The ideal use case for error.Join is if we want to do errors.Is/As when we have multiple different errors. Dont think this is ideal use case for this. Sorry for my previous comment.

aloknerurkar (approved):
LGTM! There is a small comment I added. Hopefully you can fix that before merging.

aloknerurkar (approved):

PR #580, opened 2026-04-14
refactor: replace full-node flag with explicit node-mode config

### LINKED ISSUE #5172: Investigate and refactor Bee node modes in node.go
Based on configuration option provided we have to ensure this scenarios are well implemented in Bee:

- blockchain-rpc-endpoint + swap-enable = light node with chequebook
- blockchain-rpc-endpoint = light node without chequebook
- swap-enable = error (disabled chain backend)
- neither = ultra-light node

martinconic:
Fixes https://github.com/ethersphere/bee/issues/5172. The previous approach to node mode was implicit and
inconsistent: the mode was inferred from a combination of --full-node
(bool) and whether blockchain-rpc-endpoint was non-empty, which
diverged from the documented definition of modes and allowed silent
misconfigurations.
Bee PR: [#5431](https://github.com/ethersphere/bee/pull/5431)
This PR introduces a single explicit --node-mode option with three
valid values: full, light, and ultra-light.


acud (approved):

aloknerurkar (approved):

pkg/orchestration/k8s/cluster.go
```diff
@@ -226,7 +226,7 @@ func (c *Cluster) NodeNames() (names []string) {
 // LightNodeNames returns a list of light node names
 func (c *Cluster) LightNodeNames() (names []string) {
 	for name, node := range c.Nodes() {
-		if !orchestration.Deref(node.Config().FullNode) {
+		if node.Config().IsLightNode() {
```
akrem-chabchoub, line 229: This change will exclude ultra-light nodes !
Should have another function for `UltraLightNodeNames` ?

.github/workflows/go.yml
```diff
@@ -28,9 +28,9 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@v9
         with:
-          version: v2.10.1
+          version: v2.11.3
           args: --timeout 10m
-          skip-cache: false
+          skip-cache: true
```
akrem-chabchoub, line 33: Not sure why we need `skip-cache: true` and `version: v2.11.3` in this PR ?

akrem-chabchoub (commented):
I got in bootnode logs: `"starting in light mode"`

To reproduce:
`./dist/beekeeper create bee-cluster --cluster-name=local --log-verbosity=debug`
`kubectl logs  bootnode-0-0 -n local`

.github/workflows/go.yml
martinconic: Bumped golangci-lint to v2.11.3 to match the version used in bee and fix a CI failure where v2.10.1 failed to download its remote JSON schema (golangci.v2.10.jsonschema.json context deadline exceeded). skip-cache: true ensures clean linter runs without corrupted action caches

pkg/orchestration/k8s/cluster.go
martinconic: Excluding ultra-light nodes from LightNodeNames() is intentional because checks like act and pushsync/lightnode require a blockchain-connected light node (with postage stamps/swap), which ultra-light nodes don't support.

However, adding UltraLightNodeNames() and IsUltraLightNode() makes total sense to complete the orchestration API so we can write dedicated ultra-light checks as well. I'll add that!

PR #589, opened 2026-05-18
feat(smoke): topology health probe

gacevicljubisa:
The smoke test gets `unexpected EOF` on downloads but the previous on-failure diagnostic (a topology dump of the downloader) didn't tell us which chunk is missing or why. This change turns a failure into a per-chunk report mapped to the three pushsync bugs in bee#5400.

The smoke test runs bee's chunk-splitting pipeline locally over each upload (deterministic, so addresses match what bee stores), and on a failure walks every chunk via `HEAD /chunks/{addr}` against the XOR-closest full node. `HEAD` is strict local-only, unlike `GET` which falls back to network retrieval; `Client.LocalHasChunk` wraps it.

Topology probes also run pre-upload, pre-download, and on-failure so a healthy baseline is visible. The walk is capped at 200 MB upload size, bounded at 32 concurrent HEADs, and the top 50 missing chunks are logged. Counters are exact regardless of log truncation.

### DIFF
--- a/pkg/bee/api/node.go
+++ b/pkg/bee/api/node.go
@@ -78,7 +78,11 @@ func (n *NodeService) Balances(ctx context.Context) (resp Balances, err error) {
 	return resp, err
 }
 
-// HasChunk returns true/false if node has a chunk
+// HasChunk returns true/false if node has a chunk.
+//
+// NOTE: This uses GET /chunks/{addr}, which on bee falls back to network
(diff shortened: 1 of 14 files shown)


akrem-chabchoub (approved):
Looks good, just wanted to know how those [values](https://github.com/ethersphere/beekeeper/pull/589/changes#diff-78fe9993ba5ee03cbda2f8a07dfb4dd04dc86dd006702695c4a682453843d78aR27) are chosen ?

PR #591, opened 2026-05-25
draft: add stamp-expiry beekeeper check

significance:
new integration check that buys a postage batch with low amount, uploads data, waits for the batch to expire, then asserts the batch is unusable and new uploads are rejected

### DIFF
--- a/config/local.yaml
+++ b/config/local.yaml
@@ -335,6 +335,16 @@ checks:
       postage-depth: 17
       postage-topup-amount: 100
       postage-new-depth: 18
+  ci-stamp-expiry:
+    type: stamp-expiry
+    timeout: 15m
+    options:
+      file-size: 1048576
+      postage-amount: 1000
+      postage-depth: 17
+      poll-interval: 5s
+      max-wait: 10m
+      seed: 42
   ci-stake:
     type: stake
     timeout: 5m
--- a/pkg/check/stampexpiry/stampexpiry.go
+++ b/pkg/check/stampexpiry/stampexpiry.go
@@ -1,0 +1,191 @@
+package stampexpiry
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/ethersphere/beekeeper/pkg/bee"
+	"github.com/ethersphere/beekeeper/pkg/bee/api"
+	"github.com/ethersphere/beekeeper/pkg/beekeeper"
+	"github.com/ethersphere/beekeeper/pkg/logging"
+	"github.com/ethersphere/beekeeper/pkg/orchestration"
+	"github.com/ethersphere/beekeeper/pkg/random"
+)
+
+// Options represents check options
+type Options struct {
+	FileSize      int64
+	PostageAmount int64
+	PostageDepth  uint64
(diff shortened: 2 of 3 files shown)


config/local.yaml
```diff
@@ -402,6 +335,16 @@ checks:
       postage-depth: 17
       postage-topup-amount: 100
       postage-new-depth: 18
+  ci-stamp-expiry:
```
gacevicljubisa, line 338: Better thing would be that we are spinning this kind of tests, on public testnet, as part of checks that are executed once for every RC.

pkg/check/stampexpiry/stampexpiry.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/ethersphere/beekeeper/pkg/beekeeper"
+	"github.com/ethersphere/beekeeper/pkg/logging"
+	"github.com/ethersphere/beekeeper/pkg/orchestration"
+	"github.com/ethersphere/beekeeper/pkg/random"
+)
+
+// Options represents check options
+type Options struct {
+	FileSize      int64
+	PostageAmount int64
```
gacevicljubisa, line 20: Other checks are using PostageTTL option, where you could set 24h as minimum postage validtity.

pkg/check/stampexpiry/stampexpiry.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return err
+	}
+
+	sortedNodes := cluster.FullNodeNames()
+	if len(sortedNodes) == 0 {
+		return fmt.Errorf("no nodes in cluster")
+	}
+
+	uploadNode := sortedNodes[0]
```
gacevicljubisa, line 72: Prefarable option with beekeper checks is to use radnom cluster node, instead of always using the same one. This can be achieved calling `cluster.ShuffledFullNodeClients(ctx, random.PseudoGenerator(time.Now().UnixNano()))` which returns random list of bee nodes.

pkg/check/stampexpiry/stampexpiry.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Step 1: Create postage batch with explicit amount for controlled expiry
+	c.logger.Infof("node %s: creating postage batch amount=%d depth=%d", uploadNode, o.PostageAmount, o.PostageDepth)
+	batchID, err := client.CreatePostageBatch(ctx, o.PostageAmount, o.PostageDepth, o.PostageLabel, false)
+	if err != nil {
+		return fmt.Errorf("node %s: create postage batch: %w", uploadNode, err)
+	}
+	c.logger.Infof("node %s: created batch %s", uploadNode, batchID)
+
+	// Verify batch exists and is usable
```
gacevicljubisa, line 90: There is a minimum of 10 block to wait for the batch to be usable. This would require some wait time.

significance: ok!

significance:
@gacevicljubisa totally ok with infrequent tests but would be good to have regular data and be able to a/b candidate versions! thanks for your comments!

PR #600, opened 2026-06-06
refactor: smoke test improvements

acud:
improves the metrics and readability of the smoke test


pkg/check/smoke/smoke.go
```diff
@@ -21,6 +22,17 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// result label values for the smoke metrics.
+const (
+	resultSuccess  = "success"
+	resultFailure  = "failure"
+	resultError    = "error"
```
gacevicljubisa: Do we need both? maybe just to leave the error?

gacevicljubisa (approved):
Dashboards are looking nice. Just drop the docs before merge.

martinconic (approved):

PR #623, opened 2026-09-16
feat(check): add chunk-stream check

martinconic:
Adds a `chunk-stream` check exercising bee's new `/chunks/stream` download mode (ethersphere/bee#TBD), plus the websocket client it needs.

## What it does

Streams chunks up to one node over a single websocket, then streams them back **from a different node** over another, asserting the guarantee that matters: exactly one delivery per requested address, matched by address rather than arrival order, no duplicates, no unrequested deliveries, and byte-for-byte identical data.

It also asserts that an address never uploaded comes back as `0x01` not-found — which exercises retrieval exhausting every peer, a path unit tests can't reach.

## Changes

- `pkg/bee/api/chunkstream.go` — `ChunkStreamService` on the API client. Upload and download streams, `ChunkDelivery{Status, Address, Data}`, exported status constants. Derives `ws`/`wss` from the client's scheme rather than hardcoding `ws://` as `pkg/wslistener` does, and fails the handshake if the node doesn't echo the negotiated subprotocol — which alone catches a node without the feature.
- `pkg/check/chunkstream/` — the check and its metrics, modelled on `retrieval`.
- Registered as `chunk-stream`; config in `config.yaml` and `local.yaml` (`ci-chunk-stream`).


pkg/check/chunkstream/chunkstream.go
```diff
@@ -0,0 +1,270 @@
...
+		return err
+	}
+
+	nodes := cluster.FullNodeNames()
+	if len(nodes) < 2 {
+		return fmt.Errorf("at least 2 full nodes are needed, got %d", len(nodes))
+	}
+
+	for i := range o.UploadNodeCount {
+		uploadNode := clients[nodes[i]]
```
akrem-chabchoub, line 95: When setting `upload-node-count: 3` the check will panic here in a [2 node local cluster](https://github.com/ethersphere/beekeeper/blob/b827ab28ca9188eb988bdb687c74f5aef35e98f4/config/local.yaml#L26)

akrem-chabchoub (changes requested):

martinconic:
solution changed, so closing this one

